PR #33, opened 2020-04-17
Nice defaults for spring so that initial presentation is much nicer

ForbodingAngel:
BYAR looks really nice with decent settings. With default settings it looks pretty awful. This is pretty much my bog standard springsettings to make sure that newbies are presented with a good looking game on the first run.


Ruwetuin:
unless something goes wrong, we already have a springsettings default for BAR here:
https://github.com/beyond-all-reason/BYAR-Chobby/blob/master/LuaMenu/configs/gameConfig/byar/springsettings.lua

Ruwetuin:
ok i added our springsettings as json , untested

PR #42, opened 2020-09-12
attempt to hide Singleplayer option entirely

Damgam:
🤷


gajop:
Let's not do this, I want to have SP support in the future

icexuick:
Why would you want to hide this? 
We now know how many SP/skirmishers are out there.

Damgam:
> Why would you want to hide this?
> We now know how many SP/skirmishers are out there.

Because it's not working properly and shouldn't be visible for everyone until it's fixed.

PR #69, opened 2020-12-07
Add 'Show AIOptions' toggle

rlcevg:
Depends on: https://github.com/Spring-Chobby/Chobby/pull/521


gajop:
Seems fine. OK to merge once the upstream gets merged.

Is there a way to actually test it though? I don't see any difference when testing with BAR. Does this apply only to some non-default AI?

gajop:
Disregard that - i don't think I was using it with appropriate upstream. I'll check again tomorrow

PR #103, opened 2021-04-02
Delete mapWhitelist.lua

M0nty-BAR:
This is not used.


Beherith:
Did you test that Chobby doesnt crash or throw errors if these files are missing, even if they are not used?

M0nty-BAR:
No my bad. I will need to do a test build.

M0nty-BAR:
4 Lines needed to be removed from mainConfig.lua, and then chobby would launch.

PR #106, opened 2021-04-07
Wave two of clean up

M0nty-BAR:
Cleaning up Folsom Entries. 
Capitalizing Hotstepper to fix order in Chobby.

Alphabetizing Config Entries. So when we go to audit or troubleshoot it is more straight forward.


Beherith:
i dunno, its probably good

PR #116, opened 2021-04-24
add Plethora_of_Ponds_V1 to mapdetails and add minimap+thumb for chobby

Sipke82:


M0nty-BAR:
Do you have start boxes?

Beherith:
Yes those would be nice too

Sipke82:
Could any1 help with those start boxes, i dont know what the numbers in
https://github.com/beyond-all-reason/BYAR-Chobby/blob/master/LuaMenu/configs/gameConfig/byar/savedBoxes.dat
 mean...

The map should be played top vs bottom

PR #123, opened 2021-04-29
Thumbnail minimap upgrade batch 3

M0nty-BAR:
Higher quality versions of all of the following minimaps and thumbnails


Beherith:
Thanks!

PR #126, opened 2021-05-03
disable giving commands through minimap

Jazcash:
seen a few complaints now that right click doesn't snap to location when clicking on the minimap and middle click is more awkward.

> ShoX — Yesterday at 13:28
> MiniMapFullProxy = 0 in springsettings.cfg i think it disables giving direct commands on the minimap but i dont use that anyway. can still use hotkey + command on minimap if you want to use it

### DIFF
--- a/dist_cfg/springsettings.json
+++ b/dist_cfg/springsettings.json
@@ -42,6 +42,7 @@
     "MaxTextureAtlasSizeX": 4096,
     "MaxTextureAtlasSizeZ": 4096,
     "MiddleClickScrollSpeed": -0.005,
+    "MiniMapFullProxy": 0,
     "MinimapOnLeft": 1,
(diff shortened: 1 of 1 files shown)


Beherith:
I agree that this is something users might want, but imo it breaks the UI consistency of left click select, right click issue command, middle click move camera. Just my 2c, i dont mind if this proxy goes into default mode.

Damgam:
Just make a toggle in options for this!!!

icexuick:
Does a toggle need a luaui/rules reload? or is it an easy toggle?
Anyways, i think we've heard 10x more people talking about the minimap middle-click not being that intuitive.
I'd go default = 0

FIr3baL:
@Jazcash Can you close this ?

PR #134, opened 2021-06-23
Replays delete

Born2Crawl:
Screenshots here: https://discord.com/channels/549281623154229250/549282587487502347/857016281826590761

1. Implemented manual deletion of replays. Refreshing the whole list after each delete is costly, and isn't easy to do (calling a function in a parent window from a child), so I ended up changing a map picture to a blank one for deleted replays, until the list is refreshed again.
2. Also added a check for replay file existence before playing it (the whole game was crashing before in this case).
3. I had to fix the margins on a negative_button style (red buttons in the UI) because it didn't match those of a skin.action_button, which was very visible when buttons were side-by-side.


psarkozy:
Excellent!

PR #135, opened 2021-06-23
Update replays UI for a different engine version

Born2Crawl:
1. Update option_button style to match action_button and negative_button
2. Make Delete replay button always red, regardless of replay version
3. Add a tooltip to a Play button for replays from a different engine version


Beherith:
Sorry I missed this PR, great work!

PR #140, opened 2021-08-03
Add "Battle Mode" combobox under "Adv Options"

rafidka:
# Overview

Users frequently struggle with the available battle types that can be used with "!preset" command, forget the command, or are not aware of it at all, and then they have to ask around. Having a button/combobox is a much better user experience.

## Notes

1. I noticed there is already some code that is supposed to allow the user to change the battle type. However, it only shows up for the game host (which is useless currently since the games are hosted by bots), and it doesn't correctly work actually. I think this code should be removed, but I didn't want to do so in this PR considering that I am not familiar with the code base. I added a comment on this which can be seen below.
2. I noticed some battle type-related configurations in `configuration.lua` but they seem to be incorrect/outdated. I also added a comment about this which can be found below.
3. I noticed that the "Add Team" and "Change Map" buttons have German translation as well. Even though the "Adv Options" button doesn't have translation, I thought it is better to add translation here. However, I am not sure whether the translation I used is correct as I don't know German and I simply used Google Translate :)

# Testing

I ran the game using the "Dev Lobby" configuration and covered the following cases:

1. When in single player mode, nothing shows up; this change is only for playing in servers.

image

2. When in a server, the "Battle Type" combobox shows up and correctly changes the battle type based on the selection.

image

3. I couldn't change the language to German to test the translation, so I temporarily changed the English value to have the German translation. This way I could at least confirm the word fits within the combobox boundary.


Beherith:
See : https://github.com/beyond-all-reason/BYAR-Chobby/blob/dc6aa2c540e1f136b4d0f94707b790001eb0f60d/LuaMenu/widgets/gui_battle_room_window.lua#L2010

FIr3baL:
@rafidka and @Beherith
i read over this PR from 2 years ago and it seems it was before integration of BarManager.
Closing this now, feel free to re-open or start a new one.

PR #164, opened 2021-09-23
Added welcome block about code of conduct

Teifion:


Teifion:
I've not been able to test it myself properly but I believe this is what would be needed.

Ruwetuin:
it works

PR #183, opened 2021-12-24
Adds filter for locked battles to the battle list

L-e-x-o-n:
There are a many locked rooms in the list so I thought it would be nice to have a filter for them. I copied the way it was done for the other filters, tested it and it works. (My first PR, please correct me if I made any mistakes.)


Beherith (approved):
Looks perfectly fine, if you even tested it as you say, then it should be fine. Thanks Lexon!

PR #188, opened 2022-03-16
Reduce size assets for spring launcher

p2004a:
This PR reduces the size of build package (measured on AppImage) from 304M to 143M. IMHO reducing size of launcher which then is only downloading actual game assets makes sense.


Damgam:
Oh pog. Someone fixed my lazyness.

PR #196, opened 2022-06-08
Added two new Lobby Settings

ppearson:
Added two new Lobby settings to control whether to:
1: Display the connect window on start-up
2: Whether to display the confirmation message box when clicking the Exit button.

Both of these settings default to true, so functionality as-is, should be identical to how it is currently.

For 1, I had to copy LuaMenu/widgets/gui_login_window.lua from Chobby.lua, and for 2 I specialised the code so that even if this setting enabled, pressing ESC (which is easy to do mistakenly) will still always ask for confirmation.

I had to widen the width of the checkboxes for the text titles I chose, but they still take up less width than the sliders above them.

### DIFF
--- a/LuaMenu/widgets/chobby/components/configuration.lua
+++ b/LuaMenu/widgets/chobby/components/configuration.lua
@@ -235,6 +235,7 @@ function Configuration:init()
 	self.lobbyIdleSleep = false
 	self.rememberQueuesOnStart = false
 	self.showConnectAtStartup = true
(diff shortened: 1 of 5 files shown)


FIr3baL:
hi @ppearson ,
i'm sorry that nobody answered you yet. 
Please feel free to put in more PRs.
There were a lot of changes on this topic, so i close it.
Btw, we have confirmation window.

Issue #194, opened 2022-06-02
Can't select "Two Continents" map


Sovgut:
![image](https://user-images.githubusercontent.com/26426861/171696266-0f052e80-480d-4ef6-9965-ec75f98e5ee2.png)

Nikuksis:
Yes we replaced those two maps with one Tundra Continents

PR #197, opened 2022-06-12
pull tundra continents update

Nikuksis:


icexuick:
Please add savedboxes.dat + minimapthumb.png
and test if everything works as expected in "Dev Lobby"

MasterBel2:
Tagging #194 for future reference

PR #206, opened 2022-07-01
Add experimental config with CDN to launcher

p2004a:
Final step to add experimenal CDN config option to launcher for testing once everything lands in place in other repos.


p2004a:
Moving to branch from https://github.com/beyond-all-reason/BYAR-Chobby, not in my fork

PR #231, opened 2022-08-28
Switch to CDN as default download method

p2004a:
I've already updated [pr-downloader references](https://github.com/beyond-all-reason/spring/commit/80e088c49767b8917364cc7a2ec3d68c8ed5c921) in engine, and then then [binaries in launcher](https://github.com/beyond-all-reason/spring-launcher/commit/887937649014318d0da9e6e4b67f831366ce392b). The new engine will also have [`RapidTagResolutionOrder` available](https://github.com/beyond-all-reason/spring/pull/325) that I'm setting here that resolves bug when switching between different repos.

I'm not aware of any outstanding issues.

### DIFF
--- a/dist_cfg/config.json
+++ b/dist_cfg/config.json
@@ -7,6 +7,11 @@
                 "display": "Alpha",
                 "platform": "linux"
             },
+            "env_variables": {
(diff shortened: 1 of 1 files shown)


Beherith:
Ok, this will be pushed withnew engine! :)

PR #261, opened 2022-09-09
Combine play/spectate into a single button

The-Yak:
Screenshot_274
This PR combines the play/spectate button at the cost of no longer displaying current status. Now text indicates the action executed upon press.


The-Yak:
It seems there is some desire to add que management functions here so I'm not sure if single button is workable.

PR #266, opened 2022-09-16
Update Rosetta and Twin Lakes Park Redux

The-Yak:
Startboxes PR https://github.com/beyond-all-reason/support/pull/20


The-Yak:
Put your startboxes here now: https://github.com/beyond-all-reason/spads_config_bar/blob/main/etc/mapBoxes.conf

PR #267, opened 2022-09-22
Fix possible crash when using CDN and repo updated

p2004a:
This is the known scenario I though we won't need to path when pointed by @Beherith but actually prd is called with springname, not rapid tag, so it continues being a problem.


Beherith:
Thanks for this one, quite the catch!

Beherith:
Should we be calling rapid with tag,?

p2004a:
From pr-downloader calling with tag is more optimal, but AFAIU in the protocol it's springname that is used everywhere, and tag is not available? (If I'm wrong about that, then we can easily)

It's also not possible to resolve springname to a tag without first downloading all versions... so that's a chicken or the egg problem.

PR #272, opened 2022-10-03
Intial download from archive and config URL

p2004a:
Two changes that I would like to release.

One is not controvesial: initial content of the pool is being downloaded from a weekly rebuild https://bar-springfiles-data.p2004a.com/init-pool.7z archive. The main purpose of that is cost saving from CDN and it shouldn't impact reliability of downloads.

The second one, setting `config_url` is more controversial. With this setting set, launcher on update will first download `config.json` from https://bar-config.p2004a.com/config.json and compare with existing config. If they are the same, continues, if they are not, launcher updates config.json and relaunches.

Using `config_url` allows for quicker updates of config:
- in case e.g. of engine updates no need to reinstall
- allows to e.g. create temporarily a new option in launcher with a test engine version for easier testing
- no longer need to run flatpak update to update engine (we had confused users because of that)

Disadvantages are:
- additional round-trip on start: It is quite quick, IMHO quite negligible.
- bad update of config.json on that URL can brick game install: The config update functionality will be guarded by tests that ensure it's safe. (I've not yet written that code, but it's not much: make sure json parses, title matches, config_urls are fine so it can be updated etc.

### DIFF
--- a/.github/workflows/publish_config.yml
+++ b/.github/workflows/publish_config.yml
@@ -1,0 +1,20 @@
+name: Publish launcher config.json
+on:
+  release:
+    types: [published]
+  workflow_dispatch:
+jobs:
+  publish_config:
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v3
+      - name: Publish config.json
+        run: |
+          curl --fail-with-body \
+            --request PUT \
+            --header "Content-Type: application/json" \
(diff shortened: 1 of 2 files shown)


Beherith:
i think this is a great step forward and i see no major issue with config update.

p2004a:
Ok, overall got positive feedback, not also added GitHub action to update config.json on server, so merging this.

PR #274, opened 2022-10-10
Maplist searchbox and improved loading from Zero-K

MooseOnTheLoose:
Screenshot_304

BUGS:

-Every term is searchable except Unofficial, even though Certified and Classic work


MooseOnTheLoose (commented):
Self review test

MooseOnTheLoose:
Fixed redraw. Ready to merge unless someone can figure out why Unofficial is not searchable.

Beherith:
Were object.lua and editbox.lua not present in byarchobby or different from in use by vanilla chobby? 

Regarding the unofficial search, i recall using some nasty hacks to add another invisible sorting column to the maps list

MooseOnTheLoose:
> Were object.lua and editbox.lua not present in byarchobby or different from in use by vanilla chobby?

These are present in vanilla chobby but missing functions caused errors.  These versions are from ZK chobby fork.
 
> Regarding the unofficial search, i recall using some nasty hacks to add another invisible sorting column to the maps list

What was the intended goal of these hacks? I deleted the extra column and notice no change.

PR #291, opened 2022-11-20
ONREMOVEUSER also remove user from all channels

FIr3baL:
This should fix chobby showing already disconnected users.

On chobby repo, this is not done in interface_shared.lua, which is the base of byar-chobby interface.lua

See chobby repo interface_zerok.lua function Interface:_UserDisconnected(data)
https://github.com/Spring-Chobby/Chobby/blob/33ebb37b0271726e4679c85df9f975003730ab09/libs/liblobby/lobby/interface_zerok.lua#L883-L885
Here it´s done on UserDisconnected event.

When we don´t want to touch chobby repo, best place should be here in byar-chobby´s interface.lua


FIr3baL:
This should fix chobby showing already disconnected users.

PR #298, opened 2022-11-26
Update gui_battle_room_window.lua

FIr3baL:
Background:
Number of users in teams are updated by a servermessage, which send the current state of a user.
After a user is removed from a team, the function checkRemoval will remove the team **only if it´s on top of team stack.**
So, if we get updates, that lead to an empty team, which is not on top of stack, it won´t be removed and stays empty in battle room.

Solution:
When top team of teamstack is empty and can be removed, also check next team on stack, if it can be removed.


MooseOnTheLoose:
Tested

PR #304, opened 2022-12-11
ParseBattleStatus enable processing of numbers greater than 24bit

FIr3baL:
Since Spring Lua is limited to 24 bit float numbers , greater numbers will be rounded and give random results.

This commit manually extracts bits 25-28 and then uses remaining 24bit part for extraction of lower bit battlestatus values.

### DIFF
--- a/libs/liblobby/lobby/interface.lua
+++ b/libs/liblobby/lobby/interface.lua
@@ -648,7 +648,128 @@ Interface.commands["FRIENDREQUESTLISTEND"] = Interface._OnFriendRequestListEnd
 -- Battle commands
 ------------------------
 
+
+-- if num1 > num2 return 1
+-- if num1 < num2 return 0
+-- if num1 = num2 return -1
+local function bignumStr_isGreater(num1, num2)
+	if #num1 > #num2 then
+		return 1
+	end
+	if #num1 < #num2 then
+		return 0
+	end
+  
+	local dig1, dig2
+	for i=1, #num1,1 do
+		dig1 = tonumber(string.sub(num1, i, i))
+		dig2 = tonumber(string.sub(num2, i, i))
+		if dig1 > dig2 then
+			return 1
+		elseif dig1 < dig2 then
+			return 0
+		end
+	end
+	-- equal
+	return -1
+end
+
+-- Subtraction for numbers given in string format greater than 24bits numbers (Spring Lua numbers are floats with maximum of 24bit)
+local function bignumStr_subtraction(minuid, subtrahend)
+	
+	-- calc subtrahend - minuid, if subtrahend is greater than minuid and remember sign to put it into result later
+	local sign = ""
+	local isGreater = bignumStr_isGreater(subtrahend, minuid)
+	if isGreater == 1 then
+		sign = "-"
+		local temp = minuid
+		minuid = subtrahend
+		subtrahend = temp
+	elseif isGreater == -1 then
+		return "0"
+	end
+  
+	local minuidStack = {}
+	for i=#minuid, 1, -1 do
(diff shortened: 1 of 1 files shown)


libs/liblobby/lobby/interface.lua
```diff
@@ -648,7 +648,128 @@ Interface.commands["FRIENDREQUESTLISTEND"] = Interface._OnFriendRequestListEnd
 -- Battle commands
 ------------------------
 
+
+-- if num1 > num2 return 1
+-- if num1 < num2 return 0
+-- if num1 = num2 return -1
```
p2004a, line 654: In most programming languages where there is function like this (Java, C#) it's called compare and the logic is
x < y -> -1
x == y -> 0
x > y -> 1

so it's kind of monotonic

libs/liblobby/lobby/interface.lua
```diff
@@ -648,7 +648,128 @@ Interface.commands["FRIENDREQUESTLISTEND"] = Interface._OnFriendRequestListEnd
...
+local function bignumStr_isGreater(num1, num2)
+	if #num1 > #num2 then
+		return 1
+	end
+	if #num1 < #num2 then
```
p2004a, line 659: nit: use elseif ?

libs/liblobby/lobby/interface.lua
```diff
@@ -648,7 +648,128 @@ Interface.commands["FRIENDREQUESTLISTEND"] = Interface._OnFriendRequestListEnd
...
+		return 0
+	end
+  
+	local dig1, dig2
```
p2004a, line 663: you don't need to compare digit by digit, you can just do `num1 < num2` because once you know the length is the same, you can just do lexicographical comparison on strings as numbers are sequentially in ascii.

So you end up with something like:
```
local function bignum_compare(num1, num2)
	if #num1 < #num2 then
		return -1
	elseif #num1 > #num2 then
		return 1
	elseif num1 == num2 then
		return 0
	elseif num1 < num2 then
		return -1
	else
		return 1
	end
end
```

libs/liblobby/lobby/interface.lua
```diff
@@ -648,7 +648,128 @@ Interface.commands["FRIENDREQUESTLISTEND"] = Interface._OnFriendRequestListEnd
...
+		elseif dig1 < dig2 then
+			return 0
+		end
+	end
+	-- equal
+	return -1
+end
+
+-- Subtraction for numbers given in string format greater than 24bits numbers (Spring Lua numbers are floats with maximum of 24bit)
+local function bignumStr_subtraction(minuid, subtrahend)
```
p2004a, line 678: TBH I don't exactly understand the logic here, you are using a bunch of maps, IMHO this can be much simpler:


```lua
-- works only for positive integers!
local function bignum_substract(num1, num2)
	local c = bignum_compare(num1, num2)
	sign = ""
	if c == 0 then
		return "0"
	elseif c == -1 then
		sign = "-"
		num1, num2 = num2, num1
	end
	while #num2 < #num1 do
		num2 = "0" .. num2
	end

	-- actual substraction
	local borrow = 0
	local res = ""
	for i=#num1, 1, -1 do
		local d1 = tonumber(num1:sub(i, i)) - borrow
		local d2 = tonumber(num2:sub(i, i))
		borrow = 0
		if d1 < d2 then
			d1 = d1 + 10
			borrow = 1
		end
		res = res .. (d1 - d2)
	end
	res = res:reverse()

	-- strip leading 0s
	while res:sub(1, 1) == "0" do
		res = res:sub(2, #res)
	end
	return sign .. res
end
```

PR #311, opened 2023-01-05
OnUpdateBot stop overwriting owner by nil

FIr3baL:
Maybe and old copy/paste error in OnUpdateBot

Spring protocol doesn´t provide owner nor aiLib on Servermessage **onUpdateBot**

The info about owner and ailib is provided by **AddBot** Servermessage before.

Prior implementation was overwriting already known owner and ailib on each UpdateBot Servermessage.

This fix checks if the 2 properties are already known and puts them into status. This enables OnUpdateUserBattleStatus to recognize a botupdate.


FIr3baL:
dont merge, think it´s better to fix problem at root and stop onUpdateBot send owner property at all...

FIr3baL:
1. merge PR https://github.com/Spring-Chobby/Chobby/pull/554
2. Then this

PR #322, opened 2023-01-19
Adjust shown icons in battle room to core_teams decisions

FIr3baL:
allow change values in igl_data.lua for users:
1) showRank true/false
2) showSkill true/false


Beherith:
Is this wired into the UI?

FIr3baL:
converted to draf,
i´ll include turn off skill :(
showRank = true 
showSkill = true

FIr3baL:
Watch The Fort said:
https://discord.com/channels/549281623154229250/549282587487502347/1066093436719079566
> We should show the minimum possible, we don't want an overwhelming UI
> Rank in player list - yes
> Skill number - no
> Country flag - no
> Sync status - only for those downloading

Beherith:
This is awesome!

PR #325, opened 2023-01-26
FB_AdjustBattleRoomWindowStatus

FIr3baL:
- increase padding right of skill value
- dont show ranks for spectators
- dont show sync status for spectators
- bots gets empty placeholder at skill column instead of "0"


FIr3baL:
ranks not always showingwith this commit, will edit this one next days...

PR #326, opened 2023-01-29
Fb fix sync rank changes

FIr3baL:
tested with this PR applied locally:
https://github.com/beyond-all-reason/BYAR-Chobby/pull/325

before:
- either a control for syncstatus or readywas initialized
- depended on the initial status of sync
-> so even after the status changed the control couldn´t be updated, because it wasn´t initialized before

with this fix:
- always create both controls (sync and ready)
- set visible of each according to status of sync and ready
- use same implementation at initializing user control and in battleStatusUpdate


Damgam:
YOLO

PR #339, opened 2023-03-12
fix Tetrad_V2 map size

nbusseneau:
It's actually 12x12, as can be verified when you launch a game on it.


nbusseneau:
Pinging @Nikuksis since I see you've made several commits to map list recently.

Beherith:
Ty for the PR!

PR #342, opened 2023-03-19
Don't send OnJoined and OnLeft to irrelevant channels

gajop:
Not sure why this wasn't designed as such to begin with.. Hopefully I'm not overlooking something big again :)


gajop:
Seeing as I already pushed https://github.com/Spring-Chobby/Chobby/commit/4106d08c5c47702ee1efe1e233ea3180bbe9fdda I'll just merge this as well..

PR #343, opened 2023-03-19
Add a debug script & associated notebooks for the optimization process

gajop:
I found this particularly useful when optimizing Chobby. 

You can get this kind of stuff with it:

image

image

& then work on optimizing while using identical input data.


Beherith:
Thanks, this one is awesome!

PR #345, opened 2023-03-19
Coalesce updates to the battle list that happen close to each other.

p2004a:
The cost is small delay, but it looks to effectively solve very long log in time.


p2004a:
Login with this takes ~1.2s?

Beherith:
tyvm!

PR #346, opened 2023-03-21
Control command capture and replay widget state through UI

gajop:
For capture, also write directly to file so infolog isn't polluted and so it's easier to extract desired commands

@Beherith 
Would like some eyes on this - I'm a bit unsure with how to properly initialize these widgets so they're:
1) actually disabled as desired
2) don't cause load when disabled
3) loaded at exactly the right time (after lobby and Configuration are available, but before autologin/connect, which seems to be 1.8s usually - although it's disabled now)

Also it'd be great if you have any pointers on how to have user configuration for options like `AUTO_QUIT_ON_FINISH` which seem too dangerous even for devs. Always worried I don't accidentally commit it like https://github.com/beyond-all-reason/BYAR-Chobby/commit/153619fcf359928a2bd5fbc9dd3bd3a4f37a5435 ...

### DIFF
--- a/LuaMenu/widgets/chobby/components/configuration.lua
+++ b/LuaMenu/widgets/chobby/components/configuration.lua
@@ -261,6 +261,8 @@ function Configuration:init()
 	self.showAiOptions = true
 	self.drawAtFullSpeed = false
 	self.fixFlicker = true
+	self.captureServerCommands = false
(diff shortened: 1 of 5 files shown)


FIr3baL:
@gajop and @Beherith - can we close this ?

PR #348, opened 2023-03-28
gui_community_window.lua

FIr3baL:
- deleted bullets and freeHeading, which were not used at all
- simplified it then
- walked through all items inside news and rearranged offsets
- fixed vSpacing , when timed events are used
- made vSpacing compatible with either image is used or not, time is used or not or whatever

welcomePanelItems.lua
- added next tournament event (with time)


Beherith:
Probably fine :D why draft?

FIr3baL:
> Probably fine :D why draft?

it was late for me and wanted to have 2nd look at it.
I´ll split it now with the adding of news for tournemant and push...

PR #349, opened 2023-03-31
Show Queue

FIr3baL:
- send c.battle.queue_status directly after received BATTLESAIDEX "You are now in the join-queue at position"
- interpret s.battle.queue_status and store join-queue
- seperate spec list in battle window into 2 lists, Spectators and Join-Queue
- order specs by name
- order queued-specs by queue-pos
- check for real battleStatus updates before calling listeners and only send changed parameters
- changed receivers of battleStatus updates to check for dependencies before update
- consequently respect user´s choice of lastGameSpectatorState - only change , if manually changed
- switched lobby.lua over to BYAR-Chobby


FIr3baL:
fixed 2 bugs:
1. join-queue number didnt show after rejoining a battle
2. Ready/Unready was not working at all

FIr3baL:
known smaller issue: if we don´t receive join-queue update, it may show a player with his last number in playerlist until next update arrives

FIr3baL:
> known smaller issue: if we don´t receive join-queue update, it may show a player with his last number in playerlist until next update arrives

workaround pushed: internally always overwrite queuePos=0 for users, which switch from spec to player. This way users in battle-player-list won´t ever be shown with a spec-join-queue-number.
The not refreshed join-queue will stay until we receive a queue-update from server.

PR #355, opened 2023-04-07
Add tooltips for join-queue and leave-queue and use them in non-singleplayer rooms

AntlerForce:
Also do not say $leaveq in singleplayer games.

### DIFF
--- a/LuaMenu/widgets/chobby/i18n/chililobby.lua
+++ b/LuaMenu/widgets/chobby/i18n/chililobby.lua
@@ -140,13 +140,16 @@ return {
 		rejoin = "Rejoin",
 		abandon = "Abandon",
 
-		tooltip_is_spectator = "You will spectate this game.",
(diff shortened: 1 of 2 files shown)


FIr3baL (approved):
- looks good to me
- this PR is extending Play/Spectate button reactions correctly for singleplayer games, which i never test that much, thank you :)
- tested with singleplayer and multiplayer

I´ll make some follow-up changes:
1. in multiplayer the "play"-button not necessary stands for "join-queue" - but only, when all teams are full, otherwise it´s a real play-button
2. Nothing to do with this PR, but Play button should become "active", when we are on queue, because more "play" is not possible than being in the join-queue already; currently the play button becomes active, when we really got a seat in the playerlist

FIr3baL:
And i wasn't aware, that i'm using the wrong apostroph for english all the time. On german keyboard we can access the "acute, backtick or left quote" directly, while the apostroph needs Shift+#
Lessons learned :)

PR #360, opened 2023-04-19
Keep buffering incoming commands while replaying buffer

MasterBel2:
If there are still commands in the buffer, we must play through the rest of the buffer before executing new commands, or else risk data corruption.


Beherith:
Thanks, LGTM!

PR #389, opened 2023-05-17
remove Tabula_Flooded_v04 from listed maps

nbusseneau:
It seems this map is not available server side, as selecting it in a lobby yields an error:

```
me: !map Tabula_Flooded_v04
[teh]clusterEU5[01] * Could not find matching map for "Tabula_Flooded_v04" in current map list
```

Anyway we have Tabular Remake and water level can be set to accomodate for flooding, so I guess it's fine to remove.


Beherith:
Thank you!

PR #391, opened 2023-05-22
Enable startWithTabOpen in menu to auto-open the first item (esp.for Multiplater battles)

Born2Crawl:
I find it annoying to have to click the "Battle List" button every time in Multiplayer menu, even though it's the only available option there. This PR fixes it by auto-opening the first menu item for Singleplayer (not sure about this one as there are 3 options in it), Multiplayer and Help menus -these only have one option in the menu. It also adjusts the text of the battles window when there are no rooms available, to explain that user needs to log in first.


Beherith:
Thank you, this one is amazingly good!

PR #396, opened 2023-06-04
Add player counts for teams, join queue and spectators in lobby room

wereii:
Shows total amount of players in a team (as well as join queue and spectators).

It works pretty well for normal lobbies, though just I tried this with PtaQs enormous FFA (32 players? one player per team) and when there is a lot of redraws in for the teams then the numbers seem to jump few or a lot of pixels to the right, might have to do something with team names taking more space when there is a lot of them (Team 8 vs Team 11, Team 17, etc).

image


wereii:
I tried to use a relative x position for the count label but it gets buggy for more teams then 2-3 teams, the label is shifted by a few pixels to the right with more teams, dunno why.

Beherith:
Thank you very much!

PR #409, opened 2023-06-16
Update ModOptions.lua

KyleAnthonyShepherd:
lobby-side addition of modoption of extra cortex T2 vech units


LuaMenu/configs/gameConfig/byar/ModOptions.lua
```diff
@@ -1240,6 +1240,15 @@ local options={
 		section = 'options_experimental',
 		def  = false,
 	},
+	
+	{
+		key    = 'expandedcorvecht2',
```
sprunk, line 1245: why spell "veh" with a C? won't that make it harder for people to set it

StanczakDominik (approved):

PR #410, opened 2023-06-20
Fb update get font

FIr3baL:
Apply GoogleFrogs Update from the past to configuration:GetFont
- no control is allowed to create its own font
- configuration offers a list of default fonts, which can be used by controls
- if special fonts are needed, they are instantiated by configuration to be reused again later

ComboBox
- always use objectOverrideFont
- don't initialize with itemFontSize anymore, Combobox uses objectOverrideFont for its children

EditBox
- always use objectOverrideFont
- always add objectOverrideHintFont (which is usually same than objectOverrideFont)

Deleted occurences of:
- instantiated controls with "font ="
- instantiated controls with "fontsize ="

Rewatched 3 times, found errors on each run - needs some testing...


FIr3baL:
Noticed color errors after some clicks
grafik

FIr3baL:
current issues with this PR:
- some code changes color of Configuration.font[1]
- chat fontsize is bigger than before

FIr3baL:
chat fontsize is about same as before now (except outlineColor)

To understand what happened here:
- the textbox in Console was initialized with fontsize= before.
- fontsize is deprecated - backwardcomp function resolves it to font.size in Control:New before instantiating by Object:New
- next all missing font.properties are filled from skin.lua \ general
- the merged font table is then instantiated and used as input for New Font

So, if we would like to have 100% same fonts when using objectOverrideFont, we would need to search for all Controls, that were initialized by font.size= or fontsize= and pass the properties to GetFont, that are different from our default fonts, that are:
- size (obviously)
- font = fonts/n019003l.pfb
- outlineColor = {0.05,0.05,0.05,0.9}
- shadow = true

But the most noticeable was console font so far. Maybe others are no problem to layout in general.

ToDo: Still searching for the function, that changes Configuration.font[1] to green suddenly.
Couldn't reproduce so far. Sometimes it happens after 10 sec. , sometime it doesn't at all  and sometimes after 10min.

FIr3baL:
2023-06-25 current status
most fonts are pixel-perfect the same than before
ToDo: When joining a battle with os limits, one of the Configuration default fonts is changed to red

PR #430, opened 2023-07-09
replace default !kick with !kickban

StanczakDominik:
A simple change that rewires the default kick button to use !kickban instead of !kick as per https://discord.com/channels/549281623154229250/549281623577722899/1127608504677576734

From !help:

```
!kick <username> - kicks <username> from channel, battle or game, depending on the way the command is sent to AutoHost
!kickBan <username> - kicks and bans <username> temporarily from channel, battle and game
```

Myself and @Teifion saw no use case for the first one, and the latter is usually what you want to do.

### DIFF
--- a/LuaMenu/widgets/api_user_handler.lua
+++ b/LuaMenu/widgets/api_user_handler.lua
@@ -941,7 +941,7 @@ local function GetUserControls(userName, opts)
 						if userBattleInfo and userBattleInfo.aiLib then
 							userControls.lobby:RemoveAi(userName)
 						else
-							lobby:SayBattle("!kick "..userName)
+							lobby:SayBattle("!kickban "..userName)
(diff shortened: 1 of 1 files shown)


StanczakDominik:
Hold off on merging this, please, one more thing to test...

StanczakDominik:
All right, I got kickbanned on [teh]clusterEU2[24], I was able to join clusterEU2[08], EU2[23], EU2[01], and I tested that I can join that last one to spec. Should be safe. :)

FIr3baL:
@StanczakDominik still in question here ?

StanczakDominik:
This should be good to go. I've tested the issue drivver was worried about (does a kickban on one host also apply a ban from other hosts in the host group? No).

PR #448, opened 2023-07-28
Fb add copy and refactor user drop down list

FIr3baL:
**This PR was inspired by adding "Copy User name" to user dropdown menu.
Took the chance to refactor the menu and its dependencies to match BAR's infrastructure and current command rights.**

Tech: Extended Chilly's combobox to not force any selection if not wanted.
- named it showSelection - if it's set to false, you won't see a focus when opening that same dropdown menu again
e.g.: it makes no sense to have "Change Faction" selected on next mouse click on a user

**refactored GetUserComboBoxOptions (big change) aka "user dropdown menu"**
- deleted all unused stuff
- made the code more readable in general
- merged conditions of each element to current bar's spads/server settings !!!
	this really needed update, can't count all changes, giving some examples:
	- don't show options which create a vote, if we are a spec - since we are not allowed to start votes as a spec: Change Team, Add Bonus, Force Spectator, Make Boss...
	- remove bot options, when we are not owner of bot: Remove, Change Faction
	- only show "Make Boss", if the user is not already boss
	- update comboboxoptions on 2 new events:
	1. if boss changes
	2. if we change our spectatorstatus - **because spectatorstatus changes what options we are allowed to vote for !!!**

- changed maxDropDownHeight for battle users user dropdown menu to fit the maximum amount of options without scrolling
before 300 -> 340
- added namedUserList to api_user_handler to check by name which userlist we handle right now

### DIFF
--- a/LuaMenu/widgets/api_user_handler.lua
+++ b/LuaMenu/widgets/api_user_handler.lua
@@ -31,7 +31,21 @@ local friendUsers = {}
 local friendRequestUsers = {}
 local notificationUsers = {}
 
-local clanDownloadBegun = {}
+local namedUserList = {
+	battleUsers = battleUsers,
+	tooltipUsers = tooltipUsers,
+	singleplayerUsers = singleplayerUsers,
+	channelUsers = channelUsers,
+	debriefingUsers = debriefingUsers,
+	partyUsers = partyUsers,
+	popupUsers = popupUsers,
+	statusUsers = statusUsers,
+	profileUsers = profileUsers,
+	ladderUsers = ladderUsers,
+	friendUsers = friendUsers,
+	friendRequestUsers = friendRequestUsers,
+	notificationUsers = notificationUsers,
+}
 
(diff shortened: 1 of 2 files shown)


LuaMenu/widgets/api_user_handler.lua
```diff
@@ -174,136 +190,29 @@ local function GetUserClanImage(userName, userControl)
...
-		for i = 1, #comboOptions do
-			if whitelist[comboOptions[i]] then
-				culled[#culled + 1] = comboOptions[i]
-			end
-		end
-		comboOptions = culled
-	end
+	local boss = info.battleID and control.lobby.battles[info.battleID] and control.lobby.battles[info.battleID].boss
+
+	if not (itsme or bs.aiLib) then																					comboOptions[#comboOptions + 1] = "Message" end
```
MasterBel2, line 202: Appears to be a lot of unnecessary whitespace here.

PR #451, opened 2023-08-04
dirty approach to show RunningSince for battles

FIr3baL:
store receivedTime from socket update and use it for calculating runningSince for battles and show in tooltips


FIr3baL:
This PR is not meant to be merged. It's more like a showcase, what is possible right now.

PR #468, opened 2023-09-12
modoptions: reenable `disco` anonymous mode

nbusseneau:
This was disabled in 3410c83487c03c90d83e39d97a13a21ceec32bc4 due to quirks happening with other widgets/gadgets that use `GetTeamColor`, for example `unitcloaker.lua` responsible for setting transparent cloak color on cloaked units.

We reenable it with the intent to try and address these issues over time as they are discovered by players.


nbusseneau:
Corresponding game PR: https://github.com/beyond-all-reason/Beyond-All-Reason/pull/2086

PR #469, opened 2023-09-14
Benchmark parameters update (scenario023.lua)

6AKU66:
1. Benchmark takes less to finish (4000 frame was overkill);
2. Map loading faster because this map have non-solid compression;


6AKU66:
It's my first commit to chobby, better to check if i didn't screw with something.

PR #470, opened 2023-09-14
Scenario pool update & Benchmark update + pos

Ezahiel:
Benchmark parameters update (scenario023.lua) by @6AKU66 
And Scenario pool update from 10 to 15 and Benchmark position

### DIFF
--- a/LuaMenu/configs/gameConfig/byar/scenarios/scenario023.lua
+++ b/LuaMenu/configs/gameConfig/byar/scenarios/scenario023.lua
@@ -26,7 +26,7 @@ local scenariodata = {
 	playerstarty	= "10%", -- Y position of where player comm icon should be drawn, from top left of the map
 	partime 		= 180, -- par time in seconds
 	parresources	= 1, -- par resource amount
-	difficulty		= 11, -- Percieved difficulty at 'normal' level: integer 1-10
+	difficulty		= 15, -- Percieved difficulty at 'normal' level: integer 1-10
     defaultdifficulty = "Bots", -- an entry of the difficulty table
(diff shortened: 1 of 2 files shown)


Ezahiel:
Thats how Scenario Pool will look like after and Benchmark position
Scen_win

WatchTheFort:
This is pure power creep, what is the purpose of changing the scale to 15? We need to standardize difficulty on a scale of 10.

Ezahiel:
> This is pure power creep, what is the purpose of changing the scale to 15? We need to standardize difficulty on a scale of 10.
Main purpose is to get rid of ugly look of Benchmark Scenario that Ptaq wanted to be last pos in scenario and having Scenario as 11/10 look ugly and unprofessional , so adding pool or changing other scenarios to 9/10 as max. Setting Benchmark as 10/10 wont guarantee last position on Scenario list.
Alternatively there could be separate Tab (window) for Benchamark. 
Either work so more pool or lower other scenario or add Tab for Benchmark.

Ezahiel:
@WatchTheFort unless you wanna this like this, so separately from scenario , so then pool 10/10 in scenario is fine. Benchmark would be out of scenarios window allowing to add any benchmark scenario as additional test(different settings etc)
Spring Screenshot 2023 09 15 - 22 50 43 81

WatchTheFort:
Why does this scenario exist? Is it only supposed to be available in dev mode?

6AKU66:
"Why does this scenario exist? Is it only supposed to be available in dev mode?"
People want to benchmark performance and compare with another systems.
Why it's supposed to be available only in dev mode? In another games benchmark hidden in devmodes?

PR #484, opened 2023-10-05
modoptions: rework anonymous mode descriptions

nbusseneau:
Now that descriptions are fixed in Chobby, we can rework these and make them easier to understand.

### DIFF
--- a/LuaMenu/configs/gameConfig/byar/ModOptions.lua
+++ b/LuaMenu/configs/gameConfig/byar/ModOptions.lua
@@ -1050,19 +1050,19 @@ local options={
 	},
 
 	{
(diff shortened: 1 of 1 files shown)


nbusseneau:
Corresponding BAR PR: https://github.com/beyond-all-reason/Beyond-All-Reason/pull/2123

WatchTheFort (changes requested):
Submitted feedback on the linked BYAR-Chobby PR, but doesn't changing the key names of the modoptions require simultaneously changing the widgets/gadgets?

nbusseneau:
The key names haven't been changed, I only reordered the list.

PR #528, opened 2023-12-09
cleanup server address overrides

FIr3baL:
- use local config for server address instead of forcing override
- removed old server addresses and show server change settings only if options available


Beherith:
The server3 option needs to stay, that is the integration test server :D You can add localhost if you desire

FIr3baL:
localhost is already in the list when you have placed the dev_mode.txt

FIr3baL:
Done by https://github.com/beyond-all-reason/BYAR-Chobby/pull/529

PR #555, opened 2024-01-17
Expand Username Blacklist

DeviousNull:
This PR seeks to improve the utility and scope of the existing username validation checks performed in the login window.

It does this in two stages:

In the first patch, a new Word_Library module is added, and the Login_Window class is migrated to use this module instead of implementing it itself. The goal with this migration is to facilitate the bulk addition of blacklisted strings by automatically normalizing entries. The existing blacklist is preserved, so this patch should not cause any functional/visible changes.

In the second patch, many new entries are added to the Word_Library blacklist. These are based on previously-encountered usernames which the BAR moderation team has required users to change.

### DIFF
--- a/LuaMenu/widgets/chobby/components/login_window.lua
+++ b/LuaMenu/widgets/chobby/components/login_window.lua
@@ -1,5 +1,7 @@
 LoginWindow = LCS.class{}
 
+include "LuaMenu/widgets/chobby/utilities/word_library.lua"
+
 --TODO: make this a util function, maybe even add this support to chili as a whole?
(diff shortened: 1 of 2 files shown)


Beherith:
The second existing list is specifically not very useful, as those nicks are already taken/banned :D How about reformulating the rules to catch them?

DeviousNull:
Well, I missed GitHub's notification and forgot about this for a good while, but the latest patch replaces the exact-username section with some extra entries in the other sections, which are more likely to be reused in the future.

DeviousNull:
Rebased (and retested) as requested.

Issue #572, opened 2024-02-08
Handle corrupt pool files


Beherith:
Warn user to delete and reinstall on seeing this in infolog:
```
[t=02:24:59.190571][f=-000001] Error: [PoolArchive::operator()] could not read file GZIP reason: "D:\Games\BAR\Beyond-All-Reason\data\\pool\91\b0e1b5055a56e0159522acfae8815a.gz: incorrect data check", SYSTEM reason: "Unknown error" (bytesRead=-1 fileSize=7665)
```

PR #590, opened 2024-03-07
Detect corrupt pool and push a popup

AntlerForce:
Addresses #572


WatchTheFort:
Can this be automated, to fix the issue without any user intervention?

Beherith:
> > Can this be automated, to fix the issue without any user intervention?
> 
> I don't know. There's no Lua wrapper for `FileSystem::Remove()`

os.remove should work, but I would NOT recommend doing it to the entire pool folder one by one.

Beyond-All-Reason.sdd\luaui\Widgets\gui_savegame.lua
  185: 		os.remove(pathNoExtension .. ".lua")
  188: 			os.remove(saveFilePath)

6AKU66:
When you will release this update please notify engine developers about release of this system to end-users (leave a message in #engine) since it's most likely will skew our desync rate and we need to know causing of that.

thehobojoe (approved):
PAT

PR #598, opened 2024-03-16
Add nil check before commands[1] = self.buffer .. commands[1]

AntlerForce:
Screenshot_20240315_185452


p2004a:
I don't understand how this can be possibly nil

PR #599, opened 2024-03-16
Introduce DEV modoption category

robertthepie:
Adds a new Developer Setting that when toggled:
Adds a DEV modoption category per GDT request, as well as shows HIDDEN modoptions.

# Testing Done
- [X] New Modoption categories don't show when the new setting isn't toggled, both before toggling and after untoggling
- [X] New Modoption categories do show when the setting is enabled, even if enabled when already in lobby
- [ ] Joining Multiplayer where there are already options toggled
- [ ] The new unhide options toggle setting is saved for the next time the software is opened

# Before
image

# After
image
image
image


sprunk:
so what's the difference between _DEV and _HIDDEN that warrants the different treatment? isn't _DEV already hidden as well

robertthepie:
Updated the inital comment to include the new developer setting, new after images, and moved hidden stuff to their respective category only visible when the enw setting is toggled on.
Time constraint means i don't have time to test when joining a pre-existing lobby and some other stuff.

robertthepie:
screwed up a merge conflic further, opps?

PR #601, opened 2024-03-18
Move mex to fit buildgrid

kroIya:
Metal extractors moved to buildable coordinates so they can now be upgraded to t2.


WatchTheFort:
@kroIya  Why was this closed?

PR #612, opened 2024-03-24
Load modoptions from game, instead of using a local duplicate

thehobojoe:
### Work done
The actual change is quite simple, but a great deal of the VFS stuff is completely undocumented so figuring it out took a while. The key is `VFS.UseArchive`, an undocumented feature that has a signature like this `(archiveName: string, action: Function)`. You pass a function as the second parameter, and all VFS functions *within* that call pull from the given archive.

`VFS.Include` will parse a lua file and return valid lua code (for our purposes, basically a table), and `VFS.LoadFile` will return the entire raw file as a string.

I would like some input on this approach, for basic local testing it works just fine, but I'm very unfamiliar with the overall codebase and possible ways in which this approach could fail.


Beherith:
Wow this is indeed huge! If ever in doubt about a call failing, use lua's `local success, result = pcall(func, params)`

PR #619, opened 2024-04-01
Add Tab Weights Support

robertthepie:
Adds weight support for modoptions tabs for sorting purposes.
Section weight can be added to Sections in modoptions via an (int/float) weight variable.
If a weight is not present at the time of populataing the tabs the amount of tabs is used as a negative weight to maintain the current order.
# Testing done:
- [X] Modoption tabs maintain current order if no weights present
- [X] Weights added to the modoptions file affect tab sorting


Damgam:
LGTM

PR #625, opened 2024-04-02
Fix Archive Already Loaded

robertthepie:
# Work Done
Loading modoptions now checks if the archive it wished to open for modoptions is already loaded via GetLoadedArchives, this is diffrent from the lua exposed HasArchive check which in engine makes a call to a diffrent object.
This prevents failure to open lobby if already in game of the same version.
# Test Steps
1. Open Singleplayer, normal bar game, not dev lobby.
2. Start Game
3. While Game is in the background try to join a Multiplayer Lobby
**|** With Fix, joins correctly
**|** Without multiplayer browser closes, no lobby opens


thehobojoe:
Works great for me, nice find!

PR #636, opened 2024-04-07
Allow any file extension for devmode.txt

thehobojoe:
Because windows can hide real file extensions, some people mistakenly make `devmode.txt.txt` or `devmode.rtf.txt`. This makes it so any file extension (or none, for that matter) will work.


thehobojoe:
done

sprunk:
shouldn't it have been `.txt.rtf` and not `.rtf.txt`?

PR #653, opened 2024-04-22
feat(liblobby): handle multiple newline separated commands

tetrisface:
This code aims to speed up lobby setup by letting the user submit multiple commands like battlesettings etc in a single input message.
Some samples:
image
image
image

Tested:
- [ x] standard commands
- [ x] long tweakdefs (bit of a freeze in my dev environment on paste)
- [x ] empty values
- [x] broken line copies


tetrisface:
To prevent game start crash I added support for handling tweakdefs that use "legcomlvl" 
image

PR #709, opened 2024-06-27
don't change focus when message received (give control back to user)

FIr3baL:
title says all

- Other users were able to annoy others by sending private messages. Each was triggering foreground.
- If you were typing in a private chat and press Enter it could happen, that you send it to another user, because this aggressive focus change happened.

(+ little safer implementation to iAmMentioned, copied from ZK)


FIr3baL:
@Beherith @AntlerForce
I can't see who and when this was introduced, because git history is broken here by a complete override of that file.
(In current ZK this is not existing. So, this should be from bar devs ???)

I could imagine sb. wanted to have coordinator messages always in foreground, is that possible ???
Otherwise i don't see any good reason to steal focus of users by incoming messages that heavy.

Anyway: i'm just merging now, feel free to correct me, but

PR #713, opened 2024-07-04
Decode and display first line of tweakunits/tweakdefs modoptions

DeviousNull:
Currently, players who have just joined a lobby with tweakunits/tweakdefs configured do not have an easy way to know what they do. This is sometimes alleviated by setting a Welcome message, but many players do not notice those messages. As a result, such lobbies often have a stream of new players asking "what does the tweak do?".

This PR updates gui_modoptions_panel.lua to:
- Display the length and a 4-character hash of tweakunits[N]/tweakdefs[N] values (in case tamper-detection is desired)
- If validation is successful, also display up to 25 characters from the first line of the decoded text

The following validation is performed on the configured tweakunits/tweakdefs values. If any validation fails, then only the length and 4-character hash is displayed.
- The configured text does not contain any non-Base64 characters (nothing is decoded if this is false)
- The decoded text begins with "--" (these characters are stripped and not displayed)
- The next 25 characters, or the entire first line (whichever comes first), do not contain any non-printable characters

Before:
BeforeChange

After:
AfterChange

Example Base64-encoded tweakunits string (with comment added):
[ExampleTweakunit.txt](https://github.com/user-attachments/files/16093004/ExampleTweakunit.txt)

### DIFF
--- a/LuaMenu/widgets/gui_modoptions_panel.lua
+++ b/LuaMenu/widgets/gui_modoptions_panel.lua
@@ -563,7 +563,7 @@ local function InitializeModoptionsDisplay()
 			return tweakText
 		end
 		local rawValue = Spring.Utilities.Base64Decode(value):gmatch("([^\r\n]*)[\r\n]?")():sub(1,27)
-		if not rawValue:sub(1,2) == "--" then -- First line doesn't start with a comment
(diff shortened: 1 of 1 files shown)


TheSilverHornet (approved):
Looks solid, no errors thrown even when fed with nonsense, and a huge quality of life improvement, definite upgrade on my inline tagging solution

PR #725, opened 2024-07-10
Skip comments with unprintable characters in tweakunit/def summaries.

DeviousNull:
When looking for a tweakunits/tweakdefs summary-comment and a comment with a nonprintable character is found, skip to the next line instead of giving up.

Any non-comment lines will still end processing without a summary, as before.

This change allows a single tweak string to use both a decoded summary and a "plaintext" description at the beginning of the base64-encoded string (as was sometimes done before summaries were supported).

The tweaks used for testing (and the expected results) are here: [teststrings.txt](https://github.com/user-attachments/files/16155301/teststrings.txt)

### DIFF
--- a/LuaMenu/widgets/gui_modoptions_panel.lua
+++ b/LuaMenu/widgets/gui_modoptions_panel.lua
@@ -555,23 +555,25 @@ local function InitializeModoptionsDisplay()
 		return value
 	end
 
-	local function tweakSummary(value)
-		value = tostring(value)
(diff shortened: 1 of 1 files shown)


TheSilverHornet:
Thankyou greatly, very much appreciated!

FIr3baL:
tested with above test-strings inside ZeroBraneStudio with lua 5.1 and a public base64 class.
Works as expected.

Issue #460, opened 2023-09-01
Remember options from previously hosted game


whossname:
My friend group always play co-op vs the AI. This means that every time we host a new lobby we need to remember to turn off Auto-balance. It would be great if the game remembered from our previous lobby that we don't use that option.

PeanutsRevenge:
Similar was already suggested someone else, bring able to save lobby/modoptions to a local file, that can be uploaded to set all variables.
If going so far as to remember one option, may aswell take to obvious end stage of all options.
Hopefully in the next lobby :crossed_fingers:

FIr3baL:
also see here: https://github.com/beyond-all-reason/BYAR-Chobby/issues/270

AntlerForce:
Closed via #727

Issue #270, opened 2022-10-02
[Feature] Save settings presets 


daufinsyd:
It would be time saving if we could save settings presets so that we don't need each time to enter all the settings again.

daufinsyd:
Yes, for example when you configure the chicken settings or when you set how much resources player get at the start of the party and so on :)

PeanutsRevenge:
Any kind of game.  If the settings could be saved [which aren't default] to a *.xxx file locally, then [as boss] when selected in the lobby, it posts them all, one by one.  Would be tripped up, if lobby not set to a preset (as for example if undeformable map had been enabled in the lobby, but it's not defined in the saved config, but water level is, it wouldn't disable undeformable map, which would stop water level change working), but it's better than at present.
It could become even more pressing as tweakdef mods grow, requiring multiple entries.

But, if it's already gonna be a part of chobby's successor, then, I'm sure we can soldier on.

FIr3baL:
Ok, good to know that new lobby takes this request already.
I don't know how they save it, but here is my suggestion:

Having 2 levels of presets inside chobby config:
1. classPresets for each group of presets
2. globalPresets, where each is a set of classPresets
```json
{
	"classPresets": {
		"modPresets": {
			"nameOfMy1stPreset": {
				"optionX": "valueforX",
				"optionY": "valueforY"
			},
			"nameOfMy2ndPreset": "..."
		},
		"mapPresets": {
			"mapName1": {
				"my1stPresetForMapName1": {
					"mapOptionX": "valueforX",
					"mapOptionY": "valuefoxY"
				},
				"my2ndPresetForMapName1": "..."
			},
			"mapName2": "..."
		},
		"AiPresets": {
			"BarbarianAI": {
				"MyOpBarbPreset": {
					"AiOption1": "ValueForOption1",
					"AiOption2": "..."
				},
				"TweakedBarbPreset": "..."
			}
		},
		"AutohostPresets": "..."
	},
	"globalPresets" :{
		"myDefaultConfig" : {
			"modPreset" : "nameOfMy1stPreset",
			"AIPresets" : {
				"BarbarianAI" : "MyOpBarbPreset"
			}
		}
	}
}
```
I imagine:
1. a dropdown menu inside Battle Window for choosing the global preset
2. a dropdown menu on each class of options like in modOptions window, autohost settings panel,...

Maybe that concept helps as a first step for anybody willing to code this :)

jere0500:
[...]

PR #727, opened 2024-07-15
[Feature] System for saving and loading option presets using a local file

jere0500:
Hey there, I have implemented a working version of a preset saving and loading system and would like some general feedback :). Unfortunatly, I currently keep finding bugs.

The system works both in singleplayer (Skirmish) and in multiplayer. The system allows you to save settings in your local lobby and save them to a preset, which can be restored at a later point.
The feature got requested multiple times (#270 #460).

## Implementation
In the current iteration I added a button below the advanced options to access the Preset Menu.
image

The button opens the preset loading and saving menu.
image

New presets can be selected in the drop-down menu when clicking `<new>`
image

This opens a prompt where you can enter a name.
image

### Saving
When opening the optionPreset panel, various settings are collected from the current lobby and stored in local variables.
When clicking the save button, the settings will be added to the `optionPreset.json` file under the preset name. The file will be created/ written in your installation folder.
Only options that are selected in the `Load Option` tab will be saved.

### Loading
The settings will be read from the file and saved to local variables. The values are applied at their places in the lobby. Some settings are differently applied in singleplayer and multiplayer. Additionally, you can specify the settings that should be loaded in the `Load Option` tab. If a specific setting is not defined for a preset, the setting will be ignored, and the current value will remain.

### Quirks/Workarounds
- it was necessary to expose access to the current battle lobby through multiple functions for fetching/ applying specific options.
    - Some of these might not be necessary, if there are better options available. I will be more than happy to use them
- I do not know if there is a better option to refresh elements like the drop-down menu of the preset selection:
    - I did it by deleting and recreating the elements
- Multiplayer: The modoptions set in the multiplayer lobby when joining will be cached as the default. Which allows to reset the modoptions when switching to another preset
    - (This could also be solved by getting the defaults directly)

## Features
As of now it supports saving and loading of.
- [X] modoptions (advanced options)
- [X] the currently selected map
- [X] the defined bots with their team and settings 
- [X] the defined starting rects and their position

Multiplayer specific
- [X] loading of spads related settings:
    - e.g. team, locked, etc..

### Other features:
- [X] When the `optionPreset.json` file gets damaged, it gets copied to an `optionPresetError[date].json` with the current date and replaced with a blank file.
- [X] Specific setting, what to save and load from presets:
    - [X] If an option is not defined in a set, it does not get loaded
image

## Limitations/ Known problems:
- Multiplayer: loading of the bots/ teams can lead to incorrect mapping of teams and bots and requires some moving of bots
- Presets can have a lot of options:
    - in multiplayer many options are applied with chat commands that can lead to spam (which might be unwanted)
- the json library  used in chobby can only handle compact json (new lines, tabs and spaces cause errors)
- in rare cases the presets create to many start boxes
- currently, there is a problem with saving/ loading modoptions in multiplayer not being applied correctly

### Untested:
- how loading of complex team presets work in multiplayer 
    - are people moved to teams correctly?
- that the optionPreset button is grayed out in MP (same as advanced options)
- correct functionality on Windows

## Ideas for future work/ Ideas
- implement also saving of resource boosts that can be defined when right clicking a bot
- import/ export feature of presets:
    - could be a further iteration/ implementation that allows sharing of presets (e.g. by exporting/ importing dedicated.json files, which can be shared freely)
    - this would currently require some level of json knowledge
    - could also think about a sending system

### DIFF
--- a/LuaMenu/widgets/gui_battle_room_window.lua
+++ b/LuaMenu/widgets/gui_battle_room_window.lua
@@ -1133,7 +1133,9 @@ local function SetupInfoButtonsPanel(leftInfo, rightInfo, battle, battleID, myUs
 		tooltip = tooltip,
 		OnClick = {
 			function()
-				WG.OptionpresetsPanel.ShowPresetPanel()
+				if modoptionsLoaded then
+					WG.OptionpresetsPanel.ShowPresetPanel()
+				end
 			end
 		},
 		parent = leftInfo,
--- a/LuaMenu/widgets/gui_battle_room_window.lua
+++ b/LuaMenu/widgets/gui_battle_room_window.lua
@@ -27,6 +27,7 @@ local mainWindowFunctions
 
 local showDefaultStartCheckbox = false
 local currentStartRects = {}
+local startRectValues = {} -- for exporting the raw values
 
 local singleplayerWrapper
 local multiplayerWrapper
@@ -55,6 +56,7 @@ local vote_votingsPattern = "(%d+)/(%d+).-:(%d+)"
 local vote_whoCalledPattern = "%* (.*) called a vote for command"
 local vote_mapPattern = 'set map (.*)"'
 
+local playerHandler
 --------------------------------------------------------------------------------
 --------------------------------------------------------------------------------
 -- Download management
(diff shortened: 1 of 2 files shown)


Corosauce:
I've been using this for a week now, while I can't provide feedback on the technical side for this pr as I do not have experience with this particular codebase, I do love that this exists. 

Using it for singleplayer has been great, no issues there.
Haven't used it for multiplayer so I cannot comment on the state of that.

The only suggestion I've come up with is to change the wording on the 'Cancel' button to just 'Close', as you can do things in that menu and then want to close it but it being named Cancel always makes me second guess I'm about to undo something.

AntlerForce:
I think this PR has been hung around with no progress for long enough, I have made some finishing touches and will merge as-is. If I hear too many complaints from the multiplayer side for whatever reason I will just make it appear in singleplayer (or devmode?) only.

PR #750, opened 2024-08-11
Modotpions panel: hideable suboptions, collumns, and colouring

robertthepie:
# Work Done
- Modoptions that are not their default value now highlight as red in the modoptions panel
- Modoptions can now hide or unhide other modoptions, 
> - These are them hidden in the changed modoptions panel
> - A vote to change them does not start if they have been changed and hidden
> - But they are not unset/reset if hidden, simply not shown`
- Moved the Text Labels of options to the left, as such limiting the fields to aprox. ~32 characters
- Spaghetifi the modoptions panel

# How to use:
## Hiding Options
Booleans:
- a param table `unlock`, of key string it should show when enabled
- a param table `lock`, of key strings it should hide when enabled
Lists:
- if used on lists, each item entry needs an `unlock` and `lock` table encapsulating all the options other items in the list reach
Others:
- can lock subheaders
- if multiple options hide the same option, a param `bitmask` in the parent containing a non-overlapping bitmask
## Columns
a param `column` containg a float, default: 1,
values smaller than zero moves to the left, 2 is half way across the screen, recommended values are 1 | 2, or 1 | 1.66 | 2.33
if the next item's column value is higher than the last, even if by a neglidgable amount, it will be moved to the same row
negative value places on a new row, using the absolute value to place on the x axis

# In practice
temp474
> *reload modoptions button not included


Damgam:
Pending Review/Tweaks from @robertthepie

PR #754, opened 2024-08-27
Disable "Mines, all mine" scenario

kroIya:
"Mines, all mine" in its current form offers very narrow and tedious gameplay compared to the surrounding scenarios.

This imo is not outweighed by the educational value of the scenario. Mines are not necessary to gameplay, are rarely used in pvp and never used by the AIs. 

I propose disabling the scenario until it is reworked to have more engaging gameplay, either after finishing the mission api or, if considered necessary, before that with the existing scripting tools.

### DIFF
--- a/LuaMenu/configs/gameConfig/byar/scenarios/scenario012.lua
+++ b/LuaMenu/configs/gameConfig/byar/scenarios/scenario012.lua
@@ -432,7 +432,7 @@ Scoring:
 
 }
 
-return scenariodata
+return nil -- scenariodata
 
 --[[
  [Game]


jrfondren:
It's very prominent in an ordered list, and player intuition is to play scenarios in the presented order. A quick alternative is bumping the difficulty so that it appears at the end of the list with the benchmark scenarios.

SethDGamre:
I spent 45 minutes to an hour baiting enemies into my mines.

2/10 would not recommend. Perhaps after the campaign API is finished there will be ways to issue commands and stuff so this unfun baiting mechanic can go away.

Mine practice scenario is a good idea but this one needs more time to cook

MeSaber:
Ya the scenario should reflect how you actually use mines. Luring units isnt really a thing.

Ezahiel:
IIRC i have somewhere script for this scenario for waves of enemy coming at your base, and you got to make sure to kill each wave before next and to win you need to oblitarate all waves , only issue i had IIRC there was no timer nor msgs of waves and left it behind due to work on API.

drivver44:
since ive started this ask.
i support disabling the scenario as its not good for bar. it doesnt teach anything relevant that is remotely useful to the player
it does not given any meaningful insight into things, unless you want to go into exploit territory.
scenarios taht are specifically meant to teach players should be reviewed and approved by the academy team.

PR #760, opened 2024-09-09
Custom AI profiles

rlcevg:
Allows adding new custom user profiles and blacklist undesired profiles for BARb AI from game side, without changes in `AIOptions.lua` engine-side.
(examples are in `aiCustomData.lua`: `customProfiles` and `blacklistProfiles`)


rlcevg:
If someone knows how lobby can read file.lua from a game then it can be improved by moving list of custom profiles into game repo near AI profile files. That way adding new profiles would require editing only list in game.sdd and not touch lobby.

rlcevg:
Checked `VFS.UseArchive` and it seems not viable, as it can be only used once when archive is not yet loaded. In theory it is possible to make aggregator single file `lobby.lua` in `game.sdd` that includes all relevant files from game, but it means it will load things that may not be used at all (like files for AI if player interested only in PvP).

PR #764, opened 2024-09-17
Modoptions: Changed is BAR Orange

robertthepie:
Instead of the current warning looking red, BAR orange for modoptions marked as changed.
image


icexuick:
Great!

PR #780, opened 2024-10-09
added new custom contributor flag, trans AK 47

quinnbbowman:
I guarantee the existence of this flag as a contributor option will make many people contribute so they can use it themselves


p2004a:
1. The one with the gun in my eyes is promoting violence, that's not acceptable in my eyes to find.
2. If some contributor would like to have some flag, they can propose it themselves, **for themselves**. I'm against adding flags by not contributors, not for themselves.

quinnbbowman:
I mean, I am not technically a "contributor" but I do have flag change rights due to overwatch and moderation, can i please get at least the regular trans flag merged :point_right: :point_left: :flushed:

p2004a:
Okey, then for me personally that would be fine: you will use it yourself.

But, that PR opened an active discussion about flags overall in private dev channel: right now all flags were basically geographic locations, game related, with the one the most on the edge being the pirate flag.

So: I guess we will get back here.

quinnbbowman:
understandable :)

PR #784, opened 2024-10-25
Feature/remember password option

ryanbennitt:
I wanted an option to not remember my password between sessions, partly to stop my son from connecting to multiplayer games, but also discovered the password is stored in plain text in the config file which isn't great.

Added checkbox to the register, login and settings screen to "Remember password" and added code to clear password on exit if this is not checked.

### DIFF
--- a/LuaMenu/widgets/chobby/components/configuration.lua
+++ b/LuaMenu/widgets/chobby/components/configuration.lua
@@ -76,6 +76,7 @@ function Configuration:init()
 	self.suggestedNameFromSteam = false
 	self.password = false
 	self.autoLogin = true
+	self.rememberPassword = false
 	self.uploadLogPrompt = 'Prompt'
 	self.firstLoginEver = true
 	self.canAuthenticateWithSteam = false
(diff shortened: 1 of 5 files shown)


AntlerForce:
I tested the branch and there seems to be a couple of problems:

1.  The state of the "Remember Password" checkbox is not correctly being recalled. When I uncheck the box and restart the game, the option will switch back to being checked upon the next load.
2. The actual process behind forgetting the password is tied to a single button, the Exit button in the bottom left corner of the lobby. That means if "Remember Password" is disabled and the game is exited through any other means, like pressing X on the system title bars, the password is still remembered.

ryanbennitt:
I've fixed the first point, just missed copying the field.

The second point seems to be a bigger issue, since the close button doesn't trigger the confirmation popup either. I had a look through chili docs and couldn't immediately see anything that would help. Will have a think if there's another way round this...

ryanbennitt:
Ah think I've got it, I've intercepted the password being copied prior to it being saved to file and removed the code from exit.

PR #785, opened 2024-10-27
Prevent pressing Login button repeatedly from triggering flood protection

AntlerForce:
and fix the cancel button so it returns to the flattened main menu.


AntlerForce:
lobby.connected is not valid but still appears in places https://github.com/beyond-all-reason/BYAR-Chobby/pull/433

AntlerForce:
Works now.

PR #791, opened 2024-10-30
Handle Invalid Base64 tweaks somewhat

robertthepie:
String Modoptions:
  Previosuly String modoptions couldn't be unset once set. This change makes it so that String modoptions set to a 1 character long option get hidden, and trying to set it to nothing sends through a single character to give the user an illusion of being able to unset the value. *Multiplyaer issue only

- trying to set it up to an empty string (not nil), now gets replaced with character `0` zero
- String Modoptions who's value is 1 character long now gets hidden, as the user was never able to unset

In addition the user gets warned when the base64 string for tweakdefs or tweakunits couldn't be parsed.

## Test steps:

Set either of the tweak fields to `a`
- Previously this would break the changed options panel, and maybe even crash the game on launch
- Now it will get hidden, (latter issue is game side issue)

Set either of the fields to `qqqqq`
- Now user will be warned that it couldn't be parsed

Set either of the fields to blank in multiplayer
- Previosuly a "`!bset tweaks/debug`" would go through, but have no effects due to no value being passed
- Now a value of 0 will be passed, hidden from the user giving a feeling of unsetting it, and hopefully somehow be handled by whatever processes the modoption

## Screenshots

image

### DIFF
--- a/LuaMenu/widgets/gui_modoptions_panel.lua
+++ b/LuaMenu/widgets/gui_modoptions_panel.lua
@@ -886,7 +886,12 @@ end
 				end
 				text = text .. tostring(name).. " = \255\255\255\255"
 				if (key:sub(1,10) == "tweakunits" or key:sub(1,9) == "tweakdefs") then
-					text = text .. tweakSummary(value)
+					local success, result = pcall(tweakSummary, value)
+					if success then
+						text = text .. result
+					else
+						text = text .. "\255\255\75\75".."Couldn't Parse\n".."\255\128\128\128"..shortenedValue(value)
+					end
 				else
 					text = text .. shortenedValue(value)
 				end
--- a/LuaMenu/widgets/gui_modoptions_panel.lua
+++ b/LuaMenu/widgets/gui_modoptions_panel.lua
@@ -459,6 +459,9 @@ local function ProcessStringOption(data, index)
 
 	local control
 	local oldText = localModoptions[data.key] or modoptionDefaults[data.key]
+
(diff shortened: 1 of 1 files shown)


Beherith (changes requested):
Please disambiguate boolean operator order of ... and ... or ... with parenthesis
If pcalling, please use the return error as an indicator for the user
success is double the C, double the S

ryanbennitt:
Should the text "Invalid" come from translation? e.g. i18n("invalid") assuming "invalid" en/de entries are added to i18n/chililobby.lua

robertthepie:
I believe attempting to support translation in chobby is half abbandonded, all over the place, and probably not worth it as it is to be replaced with the new lobby

> Please disambiguate boolean operator order of ... and ... or ... with parenthesis

i don't know how to do that with valid syntax

> If pcalling, please use the return error as an indicator for the user

what do i give them? `Invalid Base64`? the error message isn't game, it's just the attempt at giving some preview, typically a comment on line 1 of the tweak giving it's name/purpose

robertthepie:
Updated it somewhat to:

- allow users to fake unset string modoptions
- tell users that the base64 couldn't be parsed, in the changed modoptions preview panel

but i am just a silly little guy

robertthepie:
*this doesn't work as the game can't handle invalid base 64

drivver44:
i have tested this PR 
i do get the expected behavior.

i have only tested this pr, with no code review.

WatchTheFort (approved):

PR #453, opened 2023-08-08
Download prior engine versions on demand

FIr3baL:
This PR depends on recent spring-launcher version rollout.
https://github.com/beyond-all-reason/BYAR-Chobby/releases/tag/v1.2470.0

Enables starting replays, that were recorded with older engine versions.

Usecases:
- You made a fresh install of BAR and copied over your backuped demo files, which are from older engine versions
- You downloaded some demos from the past, e.g. tournament finals

PR #813, opened 2024-11-15
Download prior engine versions on demand: Moose 

AntlerForce:
Closes #453 (Does not require merging of original PR)

This is an expansion of @FIr3baL's PR that attempts to act upon feedback from @p2004a in regards to security and found bugs.

The engine version requested for download is now sanitized by a list of sane characters already built into chobby for username check, and adds a few more that are found in the engine label. The sanitized version is pattern-checked.

Failed download of a required engine version to run the replay is handled a bit more gracefully, the interface still isn't great as the user is requested to check the status of the download themselves and act accordingly. There is room for improvement here but things should basically work.

There is now a small notification which appears in the bottom right corner if any download fails, a small band-aid to address the lacking interface above.

### DIFF
--- a/LuaMenu/widgets/gui_steam_coop_handler.lua
+++ b/LuaMenu/widgets/gui_steam_coop_handler.lua
@@ -171,7 +171,8 @@ end
 
 -- outcome example: https://github.com/beyond-all-reason/spring/releases/download/spring_bar_%7BBAR105%7D105.1.1-1354-g72b2d55/spring_bar_.BAR105.105.1.1-1354-g72b2d55_windows-64-minimal-portable.7z
 local function GetEngineDownloadUrl(engineVersion)
-	local pureVersion = engineVersion:gsub(" BAR105", "")
+	local sanitizedEngineVersion = WG.Chobby.Configuration:SanitizeEngineVersion(engineVersion)
+	local pureVersion = sanitizedEngineVersion:gsub(" BAR105", "")
 	local baseUrl = "https://github.com/beyond-all-reason/spring/releases/download/"
 	local versionDir = "spring_bar_%7BBAR105%7D" .. pureVersion .. "/"
 	local platform64 = Platform.osFamily:lower() .. "-64"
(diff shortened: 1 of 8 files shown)


AntlerForce:
I think this is ready for review.

LuaMenu/widgets/gui_steam_coop_handler.lua
```diff
@@ -159,7 +159,30 @@ end
...
+end
+
+-- outcome example: https://github.com/beyond-all-reason/spring/releases/download/spring_bar_%7BBAR105%7D105.1.1-1354-g72b2d55/spring_bar_.BAR105.105.1.1-1354-g72b2d55_windows-64-minimal-portable.7z
+local function GetEngineDownloadUrl(engineVersion)
+	local sanitizedEngineVersion = WG.Chobby.Configuration:SanitizeEngineVersion(engineVersion)
+	local pureVersion = sanitizedEngineVersion:gsub(" BAR105", "")
+	local baseUrl = "https://github.com/beyond-all-reason/spring/releases/download/"
+	local versionDir = "spring_bar_%7BBAR105%7D" .. pureVersion .. "/"
+	local platform64 = Platform.osFamily:lower() .. "-64"
+	local fileName = "spring_bar_.BAR105." .. pureVersion .. "_" .. platform64 .. "-minimal-portable.7z"
```
p2004a, line 179: This already won't work, new engine releases won't be cut from BAR105 branch, but some other dedicated release branches.

See e.g. https://github.com/beyond-all-reason/spring/releases/tag/spring_bar_%7Bmaster%7D105.1.1-2682-g7363dbc that was cut from `master` branch.

AntlerForce: I've changed it to work with any branch name, including any that have "-" characters like https://github.com/beyond-all-reason/spring/tree/sprunk-patch-7  

but I don't think I can support any branch names that have / characters in the name, as these are already getting sanitized out. For example https://github.com/beyond-all-reason/spring/tree/badosu/allow-pure-modifiers-keysets will not work here if ever built. 

Is that a problem?

AntlerForce: https://discord.com/channels/549281623154229250/724924957074915358/1310288362623598792

This discussion indicates `rel-YYYY-MM.N` will be the likely format for branch names in the future, which is compatible with the code here now.

PR #825, opened 2024-11-25
Add a button to refresh the list of saved games and automatically refresh when a new save is detected

### LINKED ISSUE #691: Load saved game UI doesn't update
"When you're in the menu, if you click the Load button on the left it refreshes the list of savegames, but if you're already in this menu and for example you load a savegame, then during the game you make more saves and then go back from the scenario to the lobby, the UI will show the loading menu but the list of savegames is not updated so you can't see the new games you saved. You have to go back and click load again.
Yesterday I kept saving my game in a scenario but I couldn't see my savegames in the list so I restarted the scenario several times. Then I quit. When I started the game again, I saw my savegames in the list."

https://discord.com/channels/549281623154229250/1241571227328315503

AntlerForce:
Screenshot_20240395

Closes #691


AntlerForce:
The automatic refresh will need a small contribution to game repo, and as a result saved games from game versions before whenever it is added will not trigger refresh.

PR #827, opened 2024-11-28
Fix not closing sockets and bad timeout error condition checks.

saurtron:
### Work done

* Fix not closing sockets consistently at api_analytics.lua and api_spring_launcher.lua.
* Also fixes some error condition checks on client:connect calls.
  * They would result in no error be treated as an error since lua doing `not res and not res == "timeout"` becomes `not res and (not res) == "timeout"` (never true) due to operator precedence.

### Related issues

* Fixes crashing (maybe on linux only) when launcher not available (see https://github.com/Spring-Chobby/Chobby/issues/548).
* https://github.com/beyond-all-reason/spring/issues/1786

### Remarks

* You might want to test this is fine on windows before merging
  * As far as I know the api should work similarly but anyways just in case
* At least on linux this was making spring **crash quite fast** after starting a game **when not run through the launcher**.
  * It was running out of descriptors due to not closing the sockets, so after not much, it would start failing to load images, sounds...
* The bad error condition checks probably didn't have much of an effect but it meant ignoring timeout wasn't always working as intended. This might have some unintended side effect tho.
* When connect returns `res, error`, when there's an error res should always be nil and error be set to some string, still I kept the `not res and err ~= "timeout"` pattern. I think conditions could be rationally changed to `err and err ~= "timeout"` but this way feels more solid, and also I think the original developer wanted to do it like this.

### DIFF
--- a/LuaMenu/widgets/api_analytics.lua
+++ b/LuaMenu/widgets/api_analytics.lua
@@ -658,9 +658,7 @@ local function LobbyInfo()
 		local client=socket.tcp()
 		local res, err = client:connect(host, port)
 		if not res and err ~= "timeout" then  Spring.Echo("Lobby:Info Error", res, err) else client:send(message) end
-		if client then
-			client:close()
-		end
+		client:close()
 	end
 end
 
--- a/LuaMenu/widgets/api_analytics.lua
+++ b/LuaMenu/widgets/api_analytics.lua
@@ -139,11 +139,17 @@ end
 
 
 local function SocketConnect(host, port)
+	if client then
+		client:close()
+		client = nil
+	end
 	client=socket.tcp()
 	client:settimeout(0)
 	res, err = client:connect(host, port)
-	if not res and not res=="timeout" then
+	if not res and err ~= "timeout" then
 		if PRINT_DEBUG then Spring.Echo("Error in connection to Analytics server: "..err) end
+		client:close()
+		client = nil
 		return false
 	end
 	if PRINT_DEBUG then Spring.Echo("Analytics connected") end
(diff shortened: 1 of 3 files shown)


p2004a:
Nice, fixes https://github.com/Spring-Chobby/Chobby/issues/548

p2004a:
Engine can layer archives on top of other archives. In the past BYAR-Chobby was layered on top of Chobby, but then when changes become very large, all of Chobby was merged into BYAR-Chobby repo and this layering disabled.

> should I PR to them too?
Really up to you. Maybe that's not really nice of us but so far almost no fixes etc are being ported back to Chobby base. If I wanted to use base Chobby myself for something I would probably do the full diff between our Chobby and base and port stuff one by one...

saurtron:
Another thing I noticed that could use fixing is how the timeout thing is implemented. It just doesn't handle timeout reponse, sometimes the timeout can be because the server is taking some time to answer, but other times it's just because the server just isn't there. At least in my tests seem like it always gives timeout response whatever is going on, at least on api_launcher.

Also, at least on api_launcher, the current logic results in the code trying to reconnect in every Update call forever. Not very good. This should be improved to only try to reconnect every few seconds or smth.

Anyways the current PR only aims to fix the leaks and the bad timeout checks, although tbh maybe the bad timeout checks were best since then timeout was actually handled like any other error XD.

I may work on that after these PRs are resolved, but anyways just for the record.

LuaMenu/widgets/api_analytics.lua
```diff
@@ -139,11 +139,17 @@ end
 
 
 local function SocketConnect(host, port)
+	if client then
+		client:close()
+		client = nil
```
sprunk, line 144: Outdated version of patch? The upstream version doesn't have the redundant `= nil` here.

saurtron: Good catch :D, but we're upstream on this PR XD since I did it here first, I had to redo the patch for the other lobbys since this is not a fork I couldn't cherry pick, I guess I was more careful with that.

PR #863, opened 2024-12-21
Default boxes button with 1 team loads as if 2 teams

kroIya:
Maps don't have loadboxes defined for them for playing with only 1 team. There is no intended way to play with 1 team at all.

However, the "coop" preset sets the number of teams to 1 to force all the players into one team, and all the bots into the other.
This means that, while the players see 2 teams (one human and one AI), SPADS considers the number of teams to be 1 and loads (fails to load) the default boxes accordingly.

This PR forces the "Default Boxes" button to load the boxes for 2 teams if the number of teams is not greater than 1.

### DIFF
--- a/LuaMenu/widgets/gui_battle_room_window.lua
+++ b/LuaMenu/widgets/gui_battle_room_window.lua
@@ -250,10 +250,13 @@ local function SetupInfoButtonsPanel(leftInfo, rightInfo, battle, battleID, myUs
 
 				if selected == "Default Boxes" then
 					local function defaultBoxes()
+						Spring.Echo(battle.nbTeams.." here")
 						if battleLobby.name == "singleplayer" then
 							battleLobby:SelectMap(battle.mapName)
-						else
+						elseif battle.nbTeams and tonumber(battle.nbTeams) > 1 then
 							battleLobby:SayBattle("!loadboxes")
+						else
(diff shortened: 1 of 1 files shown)


AntlerForce:
> This means that, while the players see 2 teams (one human and one AI), SPADS considers the number of teams to be 1 and loads (fails to load) the default boxes accordingly.

Can you explain what you mean by "(fails to load)"? The boxes appear to load fine when I test this:

Screenshot_20240456

AntlerForce:
> > Can you explain what you mean by "(fails to load)"? The boxes appear to load fine when I test this
> 
> The maps don't have a default config for 1 team and 1 extra box, so it doesn't load the default config. It seems to load whatever was previously played on the host with that configuration, which aside from not being the default, is often unplayably skewed.

1 box + 1 extra appears to load the map's config for 2 boxes, regardless of what I have previously set as boxes on the host.

I have so far failed to reproduce the issue you are describing, where after setting coop preset the boxes loaded by `!loadboxes` and `!loadboxes \""..battle.mapName.."\" 2 0")` are not identical.

Do you have a consistent steps that conjures unplayably skewed boxes?

AntlerForce:
Related issue in maps-metadata https://github.com/beyond-all-reason/maps-metadata/issues/121

AntlerForce:
Okay thanks, the first example screenshot indeed does not match the set boxes for that map on rowy.

PR #867, opened 2024-12-23
Make minimum bonus 0 in all cases

AntlerForce:
The special -99 minimum value that the bonus selector gets in singleplayer would imply support for using bonus to debuff AI, but this value does not appear to do anything below 0. This PR sets this to 0 in all cases instead of multiplayer-only.


AntlerForce:
> Did it really not work? The handicap works in scenarios, something's not right here.

It didn't have any visible interface change in the player list, unlike the positive bonus values which did.

PR #906, opened 2025-02-01
Display Demo Time in local user time

### LINKED ISSUE #456: show local time for demos
It was 1:35 in my country, which is currently UTC+2, but it shows UTC timestamp.
grafik

Derrick633:
Resolves #456 
- Added logic to convert the demo time from UTC to Local time
  - Current State
    image
  - Timezone set to UTC +3:
    image
  - Timezone set to UTC -5:
    image

### DIFF
--- a/LuaMenu/widgets/gui_replay_handler.lua
+++ b/LuaMenu/widgets/gui_replay_handler.lua
@@ -166,6 +166,29 @@ local function playerWidget(playerInfo)
 	return ret
 end
 
+-- Convert UTC Time To User's Local Time
+local function convertUtcToLocal(year, month, day, hour, minute)
+  local utcDate = {
+    year = year,
+    month = month,
+    day = day,
+    hour = hour,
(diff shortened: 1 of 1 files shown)


AntlerForce (changes requested):
Screenshot_20240549

The way local time displays isn't consistent with other parts of the interface that do utc to local conversion, for example in my locale we use am/pm instead of 24 hour time, and this is successfully done in the welcome panel.

This is because this change uses a custom function instead of the Spring.Utilities.UtcToLocal function that already exists, see `api_timeFunctions.lua` https://github.com/beyond-all-reason/BYAR-Chobby/blob/ab9c83431f0de26371528690f9add197ef401c47/LuaMenu/widgets/api_timeFunctions.lua#L340

AntlerForce (approved):

AntlerForce:
Thanks, I'll merge after the 40k member celebration event has concluded.

PR #918, opened 2025-02-18
Spring cleaning

MasterBel2:
Cleans up some things I found while digging into the codebase. Each commit for this PR should be a non-breaking changes.


MasterBel2:
PR edited: removed reference to deleted file

PR #924, opened 2025-03-13
feat(Change Team): use drop down window

oopsbagel:
Replace the integer selection window for changing teams (a slider) with a combobox (a drop down box) for selecting teams. This should feel more intuitive and be easier to click through. This also matches the SPADS configuration interface where the number of teams and team size are organised with a combobox.

# Screenshots
## Before
change_team_old

## After
change_team_new

### DIFF
--- a/LuaMenu/widgets/api_user_handler.lua
+++ b/LuaMenu/widgets/api_user_handler.lua
@@ -1054,10 +1054,9 @@ local function GetUserControls(userName, opts)
 						if battleStatus.isSpectator then
 							return
 						end
-						WG.IntegerSelectorWindow.CreateIntegerSelectorWindow({
-							defaultValue = (battleStatus.allyNumber or 1) + 1,
-							minValue = 1,
-							maxValue = 16,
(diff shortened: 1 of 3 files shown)


oopsbagel:
I think this looks a little better with the drop down box centered?

change_team_new

oopsbagel:
force-pushed just to change the first commit to "feat".

oopsbagel:
This is ready for review, I don't speak any of the other languages in `chililobby.lua`.

AntlerForce:
> This is ready for review, I don't speak any of the other languages in `chililobby.lua`.

That's fine, this file isn't currently used for translations for BAR.

This looks good to me, thanks for the contribution.

AntlerForce:
Actually, I see font can be cropped off in some cases:

Screenshot_20240606
Screenshot_20240603

oopsbagel:
This should look better now.

simpleai
simpleconai

Changing your name is clunky enough I didn't wanna set it to TheMooseIsLoose (if I even could) but I suspect you can check it. :)

Thank you for testing this!

PR #940, opened 2025-03-29
Cleanup `gfx_limit_idle_fps.lua`

MasterBel2:
I don't know if such cleanup PRs are welcome, but I was spending so much time trying to figure out how this worked (because of vsync-related lag on my machine), I figured I"d leave behind an improvement for the next person.

Overall, functionality should be near-identical, but overall with a much more straightforward transition throughout possible states that should massively reduce unnecessary complexity that created a tonne of ambiguity when I was trying to figure it out.

This project has left more questions - most notably, how does the FPS limiting in `gfx_limit_idle_fps.lua` relate to the FPS limiting in `api_limit_fps.lua`? Maybe I'll get to that some other time.

### DIFF
--- a/LuaMenu/widgets/garbage_collection.lua
+++ b/LuaMenu/widgets/garbage_collection.lua
@@ -1,0 +1,35 @@
+
+function widget:GetInfo()
+	return {
+		name = "Garbage Collection",
+		desc = "Periodically forces garbage collection" ,
+		layer = -math.huge,
+		enabled   = true,
+	}
+end
+
+
+local basememlimit = 200000
+local garbagelimit = basememlimit -- in kilobytes, will adjust upwards as needed
+local lastGCchecktime = Spring.GetTimer()
+
+function widget:Update()
+    if Spring.DiffTimers(Spring.GetTimer(), lastGCchecktime) > 1 then
+		lastGCchecktime = Spring.GetTimer()
+		local ramuse = gcinfo()
+		--Spring.Echo("RAMUSE",ramuse)
+		if ramuse > garbagelimit then 
+			collectgarbage("collect")
+			collectgarbage("collect")
+			local notgarbagemem = gcinfo()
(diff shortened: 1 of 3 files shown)


LuaMenu/widgets/garbage_collection.lua
```diff
@@ -0,0 +1,35 @@
...
+local lastGCchecktime = Spring.GetTimer()
+
+function widget:Update()
+    if Spring.DiffTimers(Spring.GetTimer(), lastGCchecktime) > 1 then
+		lastGCchecktime = Spring.GetTimer()
+		local ramuse = gcinfo()
+		--Spring.Echo("RAMUSE",ramuse)
+		if ramuse > garbagelimit then 
+			collectgarbage("collect")
+			collectgarbage("collect")
```
WatchTheFort, line 23: This needs an explanatory comment of why `collectgarbage` needs to be called twice.

LuaMenu/widgets/garbage_collection.lua
WatchTheFort, line 1: Why do we need to handle garbage collection ourselves at all? Why does Lua's garbage collector not suffice?

WatchTheFort, line 1: Variables need to be camelCase

MasterBel2:
Since the garbage collection code is simply copy-pasted out from gfx_limit_idle_fps.lua, I’d prefer to make any changes to it in a separate PR.

LuaMenu/widgets/garbage_collection.lua
Beherith: mark + sweep

LuaMenu/widgets/garbage_collection.lua
Beherith: Because we create a ton during caching of files, and its better to manually collect than to continously run gc

PR #960, opened 2025-04-19
Add string matching for newest engine naming format

kroIya:
Since the engines using the format `branch .. "." .. engineVersion` are only:
rel2501.2025.01.3
rel2501.2025.01.5
rel2501.2025.01.6

added a separate case to match them, and the rest of the newer engines will expect `"recoil_" .. engineVersion`


A-childs-encyclopedia (approved):
tested and works, LGTM

PR #967, opened 2025-04-21
Allow battle title in tooltip to take up as many lines as needed

AntlerForce:
Screenshot_20240662


AntlerForce:
> There is currently some kind of problem with any changes to gui_tooltip causing the widget to fail to load. A new IGL_known and IGL_order must be generated to restore the tooltip.
> 
> Investigation has not yet revealed a solution: https://discord.com/channels/549281623154229250/549282587487502347/1364024576488505455

Sorted, ended up being consequence of this PR: https://github.com/beyond-all-reason/BYAR-Chobby/pull/798

PR #999, opened 2025-05-19
Hotfix: Do not ask again use value, not key

robertthepie:


LuaMenu/widgets/gui_super_announce_notifier.lua
NortySpock, line 1: Supper as in "mealtime"? I assume someone intended "super" as in "over" or "higher-than" 😉

robertthepie: 1. am dyslexic
2. I asked floris to proofread or something
3. why must you hurt me this way
4. PRs welcome

NortySpock: No worries, happens to everyone and I think you stated you were fending off a small child at the time :)

PR #1005, opened 2025-05-28
Feat: Make bset work in Singleplayer

robertthepie:
#### Test Steps:

1. Open Singleplayer Skirmish:
2. Type in in chat any bset command, some examples:

```
!bset tweakunits test
```

also includes the multiline support:

```
!bset tweakunits potato
!bset tweakdefs patata
```

Also to test:
- Going into multiplayer (straightaway or after) (and back)


saurtron (changes requested):
The bset parsing doesn't enforce good correctness rules, so I did some suggestions.

Also, what's the command to remove a modoption? (smth like `!bunset foo` maybe?) Could implement that too as part of a different PR maybe.

saurtron (changes requested):

PR #1014, opened 2025-06-21
Legion survey and new items pulse

SethDGamre:
makes it so that new items that haven't appeared before in your chobby have a subtle pulsing effect to grab your attention. After displaying once in this fashion, it's just plain white thereafter forevermore. Also allows for intentional exemptions to this pulse effect with the new `NoPulse` param

### DIFF
--- a/LuaMenu/configs/gameConfig/byar/welcomePanelItems.lua
+++ b/LuaMenu/configs/gameConfig/byar/welcomePanelItems.lua
@@ -13,6 +13,7 @@ local welcomePanelItems = {
     {
         Header = "Welcome to Beyond All Reason",
         Text = "Welcome back Commander. We hope you are ready for epic Singleplayer and Multiplayer battles. Check out our Discord and join the community!\n",
+        NoPulse = true,
     },
 
     {
--- a/LuaMenu/widgets/chobby/components/configuration.lua
+++ b/LuaMenu/widgets/chobby/components/configuration.lua
@@ -85,6 +85,7 @@ function Configuration:init()
 	self.steamLinkComplete = false
 	self.alreadySeenFactionPopup4 = false
 	self.firstBattleStarted = false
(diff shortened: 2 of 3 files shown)


SethDGamre:
I tested this out, works as it should far as I can tell. Ready for merging

robertthepie:
> don't know how to test Chobb

Similar to game, put the repo in a chobby.sdd folder next the the bar.sdd.
To launch it you select the dev lobby in the launcher.

przystuj:
OK, looks like it works as intended. https://streamable.com/b0f9pg
However, isn't 10s delay too much? I would be in the lobby already and there would be a high chance I wouldn't even notice the blinking

SethDGamre:
> OK, looks like it works as intended. https://streamable.com/b0f9pg However, isn't 10s delay too much? I would be in the lobby already and there would be a high chance I wouldn't even notice the blinking

I was testing this on my laptop, it's a lower end PC with the minimum specs required to run the game. 10 seconds was the gaurunteed time it took for my laptop to load up the game. Additionally, you kinda want that delay because when everything appears on screen all at once you need time to acclimate to what's on screen.
If things got blinky as soon as you loaded in, it'd just blend in with all the other new elements IMO

przystuj:
What if 'new' stuff had some background on the title + 'new' label alongisde the blinking? I somehow got used to 'new' things having some persistent marker until i click/hover them. In that case it could disappear on next load (as in this PR).

Or maybe it should keep blinking every minute or something.

Having a single blink after arbitrary delay feels too shy for me, considering that it's supposed to get someones attention

SethDGamre:
Visually, this is exactly what the art team boys wanted. I can pick up the conversation somewhere but I think I talked to ice and protar. 
I agree with you, but I thought that the reasoning was it would be really annoying

PR #1029, opened 2025-07-10
Computed lobby sizes

wrath625:
Hi all, first contribution.  Hoping I'm doing it right.

This change alters the lobby size text to show a computed value of how many players the lobby can support based on battle.teamSize * battle.nbTeams.

The value is clamped to maxPlayers to avoid having a player set a lobby larger than 16 (8 teams of 4 or something), but still allows for developers to set a much larger lobby than the regular 16 and have it still display correctly.

Example:
image

### DIFF
--- a/LuaMenu/widgets/chobby/components/battle/battle_list_window.lua
+++ b/LuaMenu/widgets/chobby/components/battle/battle_list_window.lua
@@ -691,9 +691,12 @@ function BattleListWindow:MakeJoinBattle(battleID, battle)
 		align = "right",
 		valign = 'center',
 		objectOverrideFont = myFont2,
-		caption = lobby:GetBattlePlayerCount(battleID) .. "/" .. battle.maxPlayers,
+		caption = lobby:GetPlayerOccupancy(battleID),
 		parent = parentButton,
 	}
+
+	parentButton.previousMaxPlayers = lobby:GetBattleMaxPlayers(battleID)
+	parentButton.previousPlayerCount = lobby:GetBattlePlayerCount(battleID)
 
 	return parentButton
 end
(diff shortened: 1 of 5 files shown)


geekingfrog:
Yeah, chobby must be able to cope with the server not returning this info. I want to put it behind a feature flag, so it may get turned off at any time.

wrath625:
I updated this so that nothing fails if the backend doesn't broadcast the team information.  With the backend using a feature flag should another one be mirrored here? Is that a thing?

RogShotz:
So whats the status on this, had a player asking for this exactly recently. Reminded me to check :)

wrath625:
> So whats the status on this, had a player asking for this exactly recently. Reminded me to check :)

It needs this PR: https://github.com/beyond-all-reason/teiserver/pull/705 for the new backend broadcast info.  Its close!

geekingfrog:
There are several syntax errors on these changes, namely, missing closing `)`. Starting chobby on this branch give errors immediately.
I've attached a git patch to fix these issues: `git patch apply /path/to/patch/file`  With the patch applied and the server enabled it works as expected.

[chobby.patch.txt](https://github.com/user-attachments/files/21546953/chobby.patch.txt)

wrath625:
Fixed this.  I got out of sync locally (fixed that too lol)

I think even if the backend is turned off it will be okay to only see a true lobby size for only lobbies you join.  There's a case where it would become stale/out of sync if you joined a lobby, left, and then that lobby changed the teamSize or nbTeams value - but I think that's pretty unlikely.

FIr3baL:
I revieved this and tested it.
It's working as expected.

Though i'd like to refactor this a little to avoid duplication. See my PR on your Branch.

PR #1034, opened 2025-07-26
Add rejoin battle button

RogShotz:
Add a button that allows users to rejoin a battle if they did not properly leave a battle, i.e. press leave battle, or leave lobby.

This allows users to reconnect to lobbies that are open after a break or to allow users to rejoin a server in the event of a crash. No longer do people have to keep a mental note of what game they were in.

To note, a known bug exists wherein if a BattleLobby opens up with the same the same ID as the last BattleLobby when it was closed, the user will connect to that 'random' lobby. This bug ultimately is very small though.

### DIFF
--- a/LuaMenu/widgets/gui_battle_rejoin.lua
+++ b/LuaMenu/widgets/gui_battle_rejoin.lua
@@ -14,38 +14,48 @@ end
 
 --------------------------------------------------------------------------------
 --------------------------------------------------------------------------------
+local lobby, battleList
+
+local function LoginRejoinOption(listener, battleID)
+	-- Check if a battle ID has been stored for rejoining.
+	if not WG.Chobby.Configuration.rejoinBattleID then
+		return
+	end
+	-- Spring.Echo("Saved BattleID: "..WG.Chobby.Configuration.rejoinID)
+
+	local battle = lobby:GetBattle(WG.Chobby.Configuration.rejoinBattleID) -- Battle Object
+	if battle == nil then
+		return
+	end
+	
+	local function doJoinBattle()
+		battleList:JoinBattle(battle)
+	end
+	
+	WG.Chobby.ConfirmationPopup(doJoinBattle, i18n("rejoinBattlePopup"), nil, 315, 200, i18n("rejoin"), i18n("abandon"))
+end
+
+local function OnJoinedBattle(listener, battleID, userName)
+	if userName ~= lobby:GetMyUserName() then
+		return
+	end
+
+	-- Spring.Echo('Saving Lobby ID: '..battleID)
+	WG.Chobby.Configuration:SetConfigValue("rejoinBattleID", battleID)
+end
+
+local function OnLeftBattle(listener, battleID, userName)
+	if userName ~= lobby:GetMyUserName() then
+		return
+	end
+
+	-- Spring.Echo('Removing Lobby ID: ' .. battleID)
+	WG.Chobby.Configuration:SetConfigValue("rejoinBattleID", nil)
+end
+
 local function DelayedInitialize()
-	local lobby = WG.LibLobby.lobby
-	local battleList = WG.Chobby.BattleListWindow()
-
-	--Spring.Echo("Saved BattleID: "..WG.Chobby.Configuration.rejoinID)
-
-	local function LoginRejoinOption(listener, battleID)
-		-- if no ID is present then there is no server to rejoin
-		if not WG.Chobby.Configuration.rejoinID or WG.Chobby.Configuration.rejoinID == 'nil' then
-			return
-		end
-
-		local battle = lobby:GetBattle(WG.Chobby.Configuration.rejoinID) -- Battle Object
-		if battle == nil or battle == 'nil' then return end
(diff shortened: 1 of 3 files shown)


LuaMenu/widgets/gui_battle_rejoin.lua
```diff
@@ -0,0 +1,62 @@
...
+		if userName ~= lobby:GetMyUserName() then return end
+
+		--Spring.Echo('Removing Lobby ID: '..battleID)
+		WG.Chobby.Configuration:SetConfigValue("rejoinID", nil)
+	end
+
+	lobby:AddListener("OnLoginInfoEnd", LoginRejoinOption)
+	lobby:AddListener("OnJoinedBattle", OnJoinedBattle)
+	lobby:AddListener("OnLeftBattle", OnLeftBattle)
+end
```
FIr3baL, line 53:
```suggestion
local lobby, battleList

local function LoginRejoinOption(listener, battleID)
	-- Check if a battle ID has been stored for rejoining.
	if not WG.Chobby.Configuration.rejoinBattleID then
		return
	end
	-- Spring.Echo("Saved BattleID: "..WG.Chobby.Configuration.rejoinID)

	local battle = lobby:GetBattle(WG.Chobby.Configuration.rejoinBattleID) -- Battle Object
	if battle == nil then
		return
	end
	
	local function doJoinBattle()
		battleList:JoinBattle(battle)
	end
	
	WG.Chobby.ConfirmationPopup(doJoinBattle, i18n("rejoinBattlePopup"), nil, 315, 200, i18n("rejoin"), i18n("abandon"))
end

local function OnJoinedBattle(listener, battleID, userName)
	if userName ~= lobby:GetMyUserName() then
		return
	end

	-- Spring.Echo('Saving Lobby ID: '..battleID)
	WG.Chobby.Configuration:SetConfigValue("rejoinBattleID", battleID)
end

local function OnLeftBattle(listener, battleID, userName)
	if userName ~= lobby:GetMyUserName() then
		return
	end

	-- Spring.Echo('Removing Lobby ID: ' .. battleID)
	WG.Chobby.Configuration:SetConfigValue("rejoinBattleID", nil)
end

local function DelayedInitialize()
	lobby = WG.LibLobby.lobby
	battleList = WG.Chobby.BattleListWindow()

	lobby:AddListener("OnLoginInfoEnd", LoginRejoinOption)
	lobby:AddListener("OnJoinedBattle", OnJoinedBattle)
	lobby:AddListener("OnLeftBattle", OnLeftBattle)
end
```
Suggestions:
1. Reduce indentation for readability
2. Instead of an anonymous function, you could implement it as a normal function for better readability
3. Though i18n is forced to "en" for bar at the moment, we should use it for all new implementations. If anybody has the time, we don't make it more work as it is already. Add this to LuaMenu\widgets\chobby\i18n\chililobby.lua in the end of "en"-section:
```
-- Battle Rejoin
	rejoinBattlePopup = "You were in a lobby last time. Rejoin it?",
```
I  suggest to shorten the term.
("rejoin" and "abandon" already exist over there)
4. Though some files in chobby make heavy use of inline code, i suggest du spread our if-statements over multiple lines for better readability. I know there is no consistency in general.
5. Use above suggestion for rejoinBattleID renaming
6. No need for checks on string literal "nil", if the values are either nil | table or nil | number

FIr3baL (commented):
All of my comments are just personal suggestions for improvement. I'm contributing as a community member, not as an official BAR developer, so feel free to treat them as optional.

1. When clicking "abandon", the stored value for rejoinBattleID should be cleared.
Otherwise, the popup may reappear on the next login — for example, when the user logs out and back in manually, or after a reconnect to the lobby server.

2. Additional use case:
If the user is in a lobby and it gets closed (BATTLECLOSED), we could reuse this widget for that scenario as well.
This could happen, for example, due to a !rehost command or a crash of the spads autohost instance.
I suggest adding a hook for OnBattleOpened, such as:
```lua
OnBattleOpened(listener, battleID, battle)
-- check if this is our stored battleID and ask for rejoin
```

3. When the game is exited normally using the "Exit" button, the question arises whether the stored rejoinBattleID should also be cleared.
I believe it should.
The important part here is to distinguish whether the exit was triggered manually or due to an error.
I suggest adding the following line at this point in the code:
https://github.com/beyond-all-reason/BYAR-Chobby/blob/71018c27a7d3aa08b4902843613eec9d3dd34218/LuaMenu/widgets/chobby/components/interface_root.lua#L387
```lua
WG.Chobby.Configuration:SetConfigValue("rejoinBattleID", nil)
```

RogShotz:
> Very nice! I can't really comment on the code itself as I'm not very familiar with it. However you might need a mechanism that will also ignore "stale" rejoin IDs. 
> 
> After a certain amount of time it's unlikely for the game to still exist. After 4 hours the battle is over I would assume and you can ignore the rejoin maybe.

I agree, it is the bug outlined at the end of this PR. I figure though hey, if a lobby is still pumping though maybe theyd want to rejoin it. If it is stale and no lobby exists with that ID though it does not pop up though :)

RogShotz:
Thank you @FIr3baL all of your comments were great and I agree with most of them. And most were implemented :)

PR #1041, opened 2025-08-19
Add Modes

### LINKED ISSUE #1040: Sharing Options
# Mode First, Details Second

## Premise
New **Sharing** tab where a **mode** at the top sets/locks/hides the **details** below. The UI orchestrates; gadgets enforce.

## UI

* **Mode dropdown** at top.
* **Functional options** below (Units, Resources, Allied Construction/Capture, Reclaim, T2 Mex/Geo Upgrades).
* Modes are declarative: they set values, lock/unlock, and hide/show rows.

##Why this?

* **Composable:** Add a mode or a feature without touching others.
* **Separable concerns:** Clean contract between UI and gadgets.
* **Safe iteration:** Defaults + locks make experiments predictable.

## Example modes (one-liners)

* **No Sharing:** All sharing off; 30% tax; most controls locked (Capture stays Automation-restricted).
* **Limited Sharing:** T2 constructors only; tax adjustable; metal send threshold 440m; tight locks elsewhere.
* **Enabled:** Everything on with fixed defaults (locked).
* **Customize:** Everything editable.

keithharvey:
**Depends on beyond-all-reason/Beyond-All-Reason#9113** (the game axis module): this lobby reads the `modes.json` that PR exports and the `values.lua` it ships. Merge that first; without it this PR falls back to the plain modoptions panel.

Closes #1040

## Summary
Loads the game's declared modes (beyond-all-reason/Beyond-All-Reason#9113) into the lobby. The SPADS plugin that applies the same modes on the host is https://github.com/beyond-all-reason/spads_config_bar/pull/228.

### Modes in the lobby
A mode is declarative Lua the game defines: a preset that picks a category's options and says which are shown, hidden or locked. Every category that ships presets gets a mode dropdown; the panel then shows exactly what the preset claims and nothing else. The lobby reads the presets out of the game archive, so client and game can never disagree, and it reads the game's value formatter from the same archive, so the strings it compares and sends are the ones the game baked. An archive that ships modes without that formatter is treated as shipping no modes, with an error logged, rather than carrying a copy here that could drift.

A clean mode pick sends only the mode key; the host expands it. What the user changes on top of the preset is sent as deviations, and locked options are never sent. Whether a mode may count for rating is part of its preset too: an unranked mode, or one that never says so, pins ranked_game off like any other claim.

### Game tab
The game tab gets a mode dropdown for the game axis: Standard, FFA, Team FFA, Territorial Domination.
image

### Transfer tab
The transfer axis gets its own tab, with the presets the transfer module declares (beyond-all-reason/Beyond-All-Reason#8521): Enabled, Disabled, Easy Tax, Customize, Tech Core. Same mechanism, one more category.
image

## LLM Disclosure
Tons of it. Most recently Fable 5.1. I have read every line multiple times at this point -- this is what I would write myself.


LuaMenu/widgets/gui_modoptions_panel.lua
```diff
@@ -655,11 +815,250 @@ local function PopulateTab(options)
...
+		local allowRanked = (mode.allowRanked ~= false)
+		WG.ModePolicy = WG.ModePolicy or {}
+		WG.ModePolicy[category] = WG.ModePolicy[category] or {}
+		WG.ModePolicy[category].allowRanked = allowRanked
+		WG.ModePolicy[category].modeLocked = {}
+		rankedBadge:SetCaption(allowRanked and "" or "Not Ranked")
+
+		isProgrammaticUpdate = true
+		if not allowRanked then
+			localModoptions["ranked_game"] = "0"
```
burnhamrobertp, line 899: When a mode has allowRanked false this forces ranked_game to '0', but I don't see anywhere that sets it back when you switch to a mode that does allow ranked. So if I pick a non-ranked mode and then switch back to a ranked one, won't ranked_game stay at '0'? And since it goes out through SetModOptions as a normal value change, I think it actually gets sent to the host, not just shown locally. It's also easy to miss because SetRankedModeAllowed just below doesn't seem to be defined anywhere, so nothing updates the battle-room UI. Could this restore ranked_game on the allowRanked path?

LuaMenu/widgets/gui_modoptions_panel.lua
```diff
@@ -655,11 +815,250 @@ local function PopulateTab(options)
...
+					rowData.x = rowData.x + (column - 1) * 625
+					row = row + 1
+					rowData.rowOrginal = rowData.y
+					modeScroll:AddChild(rowData)
+				end
+			end
+		end
+
+		if mode.modOptions then
+			for optKey, rule in pairs(mode.modOptions) do
```
burnhamrobertp, line 1061: This only sets or clears lockedOptions for keys that are in the new mode's modOptions, and lockedOptions itself looks like it's only reset when the window is first created, not here. So if mode A locks an option and you then switch to a Customize/retain mode that doesn't list it, won't that key keep its old lockedOptions entry and then get overwritten with the server value on Accept? I think someone who had set that option would see it silently revert. Would resetting lockedOptions at the top of applyMode, the same way modeLocked is rebuilt, fix it?

spads-plugins/ModeCommand/modecommand.py
```diff
@@ -0,0 +1,255 @@
...
+
+    if len(params) < 2:
+        spads.invalidSyntax(user, 'mode')
+        return 0
+
+    category = params[0]
+    mode_key = params[1]
+    kv_params = params[2:]
+
+    category_data = _as_dict(_modes().get('categories')).get(category)
```
burnhamrobertp, line 193: This calls .get('categories') straight on whatever _modes() returns. If the fetched or cached json is ever not an object (a list, say), won't that throw an AttributeError right into the perl bridge? The accesses below it go through _as_dict but this first one doesn't. Could the top level be wrapped the same way?

spads-plugins/ModeCommand/modecommand.py
```diff
@@ -0,0 +1,255 @@
...
+            return 0
+        settings.append((m.group(1).lower(), m.group(2)))
+
+    # Explicit key=value params may not override options the mode marks as locked.
+    locked_keys = set(
+        key for key, spec in mod_options.items()
+        if isinstance(spec, dict) and spec.get('locked')
+    )
+    for (key, _val) in settings:
+        if key in locked_keys:
```
burnhamrobertp: The user's key gets lowercased up at line 209, but locked_keys is built from the mod_options keys as-is. So if a mode locks 'ResourceMult' and someone passes resourcemult=something, won't this check miss it and apply the override anyway? I think BAR option names are all lowercase in practice so it may not bite today, but it makes the lock weaker than it looks. Lowercasing both sides would close it.

spads-plugins/ModeCommand/modecommand.py
```diff
@@ -0,0 +1,255 @@
...
+    base = _release_base()
+    urls = []
+    slug = _version_slug(mod_name)
+    if slug:
+        urls.append('%s/game_modes-%s.json' % (base, slug))
+    urls.append('%s/game_modes-%s.json' % (base, _channel(mod_name)))
+
+    for url in urls:
+        try:
+            with urllib.request.urlopen(url, timeout=_FETCH_TIMEOUT) as resp:
```
burnhamrobertp, line 107: This is a synchronous urlopen on what I understand is SPADS's single thread, with a 5s timeout and up to two URLs, so a slow or unreachable release could stall the host for around 10s on the first !mode after a rehost. From what I can see the vote-eligibility check reaches it too, which a normal player can trigger. It's once per version so not constant, but a multi-second freeze of the autohost still seems bad. Could this be backgrounded or moved to load time?

burnhamrobertp:
I'm not deep in the SPADS side or a lot of this Lua, so some of this might be me misreading it.

1. Fetching game_modes.json over the network at runtime. The plugin pulls it from a GitHub release the first time someone runs !mode after a rehost. Is there a reason it has to be fetched at runtime instead of shipped inside the game archive the host already has synced? I get that the host can't regenerate it itself since it's baked in CI from the modes lua, but if CI is already producing the file, couldn't it be bundled into the archive so the host just reads it locally? My worry with the release fetch is that the failure modes are all silent: if the asset is missing, or the host's network is locked down, or the slug doesn't match, the host just quietly serves stale or empty data, which seems painful to debug for whoever is running the host.

2. How do we guarantee the host's modes match the version players are actually on? From what I can tell the right file is picked by slugifying the game's display name, and the schemaVersion field in the json isn't read by anything. If the pinned asset isn't there it falls back to a rolling channel file that isn't tied to the hosted version. Won't that let the host apply one version's settings while clients read a different version's modes out of the archive, with nothing flagging the mismatch? Even just checking schemaVersion and logging loudly when it falls back would help, I think.

3. The expansion logic lives in two places. The rule that turns a mode into its full set of modoptions is in the Lua resolver (moderesolver.lua) and again in the CI generator that bakes the json, in two different languages, and I don't see anything testing that they agree. Since the drift between copies looks like what caused several of the regressions this PR is already fixing, would a shared fixture or a test that runs both and compares be worth it? Otherwise I think it just drifts again the next time someone touches one side.

keithharvey:
@burnhamrobertp Man those were great catches. I apologize for the lapse in reply. Ironically, I have been working in there and reviewing this myself ignorant of your review. I'm going to go over your comments with a more fine toothed comb now but I do think I addressed a lot of these already.

Re your concerns for my quick hot takes:

1) This is purely to avoid talking to Marek, it pains me. I am sure with a little more time I could find a way to get my JSON file from game -> spads plugin IN infrastructure code (correct), with a path to also do that locally. That said, I don't love the blocking but yeah, I stuck my hand in the[ infrastructure hornets nest](https://github.com/beyond-all-reason/ansible-spads-setup/pull/79) and would rather not go there to get my json file in the directory. This spike is big enough already. Aaaaaand the client is theoretically dead. People tell me it's dead. I don't necessarily believe them but if this gets merged we have a master version of game_modes.json and we can go over to the new client keep symmetry easily enough.
2) It's pretty fregin hokey because bar release versioning is kind of hokey, especially true in the Chobby runtime. Let me get back to you on this one with more detail. I know this was an imperfect solution relying on "stable" = god, with customization of branches allowed by publishing an individual commit SHA on explicit running of an action in BAR. That seemed like enough for now since these are unlikely to change fast. If yall let me support devops during the first release after publication, we're going to be gucci.
3) This was a great callout, I have since fixed this. I would think adding busted here (and unit tests on the munging and unpacking in particular) makes a ton of sense. That said, I am running low on mana right now. I need to merge these spikes then look to the future. I'm carrying too much code rn. That said, I would do it as an independent PR if you can review/merge it. :dancers:

LuaMenu/widgets/gui_modoptions_panel.lua
keithharvey: Real. The prior ranked_game value is now remembered when switching to an unranked mode; the pin holds while any category's mode is unranked and lifts when none is, restoring it. Removed the battle room window SetRankedModeAllowed since it assumed the room had a ranked indicator to update. I did have that at one point, but have since removed it.

LuaMenu/widgets/gui_modoptions_panel.lua
keithharvey: Good catch. applyMode now releases the locks the category placed last time before it rebuilds them, skipping any the new mode also locks so a key locked by both never flickers. Accept was indeed treating the stale entry as a host lock and reverting the value to the server's.

spads-plugins/ModeCommand/modecommand.py
keithharvey: Both sides lowercase now.

spads-plugins/ModeCommand/modecommand.py
keithharvey: Yep. All dict all the time now.

spads-plugins/ModeCommand/modecommand.py
keithharvey: Moved this into a worker thread that only touches the network and the cache file, and the plugin warms it at load so the first `!mode` after a rehost normally finds it already there. If it hasn't landed yet, `!mode` serves the last cached copy and logs that it did. The worker never calls into spads. The main thread emits its log lines when it collects the result.

keithharvey:
@burnhamrobertp back to your three, now that I've been through the inline ones (all five addressed above):

1) Runtime fetch: unchanged, for the reasons in my last reply. It stays a release asset until the plugin can read it out of the archive with infra on board.

2) Version match: the fallback is loud now. If there's no asset pinned to the hosted commit's SHA and it drops to the rolling channel file, that logs a warning naming the SHA. schemaVersion is read, and a file the plugin doesn't understand logs an error and says so. Still keyed on the SHA in the version string for the pinned case, still "stable" means stable otherwise.

3) You were right that it would drift again. The expansion was already baked once by the game's export in CI, but the value formatter was copied into the lobby with a "must match" comment and nothing testing that it did. It's now one function in the game (modules/modes/lib/values.lua, #8486) that the export uses and the lobby includes out of the game archive when it loads the modes. There's no copy left on this side.

Mind taking another look?

LuaMenu/widgets/gui_modoptions_panel.lua
```diff
@@ -1094,6 +1766,29 @@ local function InitializeModoptionsDisplay()
...
 
+		-- Reflect the mode chosen in the battle (e.g. sharing_mode) so the tab shows
+		-- the active mode even when it was changed externally (SPADS, other players).
+		-- If the modoptions window is open, live-refresh that category's panel so the
+		-- selector and its sub-options update without reopening.
+		if activeModes and not applyingModes then
+			for cat in pairs(activeModes) do
+				local battleKey = panelModoptions[cat .. "_mode"]
+				if battleKey and selectedModeKeys[cat] ~= battleKey then
+					selectedModeKeys[cat] = battleKey
```
burnhamrobertp: This lands before applyMode reads the same slot as previousModeKey, so `previousModeKey ~= modeKey` comes out false and the reset-to-defaults loop gets skipped. On a mode change arriving from the host, wouldn't the new preset then be applied over the previous mode's values, leaving anything the new mode doesn't claim wherever the old one set it?

LuaMenu/widgets/gui_modoptions_panel.lua
```diff
@@ -655,11 +832,331 @@ local function PopulateTab(options)
...
+				else
+					lockedOptions[optKey] = nil
+					WG.ModePolicy[category].modeLocked[optKey] = nil
+					SetControlLock(optKey, false)
+				end
+			end
+		end
+	end
+
+	local defaultSelected = 1
```
burnhamrobertp, line 1094: BAR's game modes are standard, ffa, team_ffa and territorial_domination, nothing keyed `enabled`, so when selectedModeKeys[category] is nil this falls through to index 1 (ffa.lua in DirList order) and Accept sends `!mode game ffa`. game_mode already declares `def = "standard"`, could that be the fallback instead?

LuaMenu/widgets/gui_modoptions_panel.lua
```diff
@@ -655,11 +832,331 @@ local function PopulateTab(options)
...
+						and opt.type ~= "section" then
+					addRow(opt)
+				end
+			end
+		end
+
+		if mode.modOptions then
+			for optKey, rule in pairs(mode.modOptions) do
+				if rule.locked then
+					lockedOptions[optKey] = 1
```
burnhamrobertp: Every deathmode item carries lock/unlock lists for territorial_domination_config and territorial_domination_elimination_threshold_multiplier, and game.end emits those same two keys with locked = false. postLock drains after CreateModePanel has already run applyMode, so processChildrenLocks shifts the row by 4095, then a later mode switch clears lockedOptions here without putting it back. Am I misreading, or does that leave the row sitting offscreen while lockedOptions says it's unlocked?

LuaMenu/widgets/gui_modoptions_panel.lua
```diff
@@ -655,11 +832,331 @@ local function PopulateTab(options)
...
+				rowData = ProcessListOption(opt, row)
+			elseif opt.type == "separator" then
+				rowData = ProcessLineSeparator(opt, row)
+				row = row - 0.5
+			end
+			if rowData then
+				column = math.abs(opt.column or 1)
+				rowData.x = rowData.x + (column - 1) * 625
+				row = row + 1
+				rowData.rowOrginal = rowData.y
```
burnhamrobertp, line 1017: PopulateTab stamps filterText/filterType/optionKey/optionData at this point and applyModoptionFilter skips any child whose filterText is nil, plus it walks tab.contentPanel.children, which for a mode tab is parentPanel rather than modeScroll. Between the two, wouldn't every option in options_main drop out of the search box once it folds into the game panel?

LuaMenu/widgets/gui_modoptions_panel.lua
```diff
@@ -1287,14 +1991,94 @@ function ModoptionsPanel.LoadModoptions(gameName, newBattleLobby, forceReload)
...
 
+	local function LoadModes()
+		local byCategory = {}
+		-- The value formatter is the game's, read from the same archive as the
+		-- modes, so what this lobby sends is what the game's export baked.
+		local okValues, values = pcall(VFS.Include, "modules/modes/lib/values.lua", nil, VFS.ZIP)
+		if okValues and type(values) == "table" and values.ToModOption then
+			ModeResolver.UseValues(values)
+		else
+			Spring.Log(LOG_SECTION, LOG.ERROR, "game archive ships modes but no modules/modes/lib/values.lua: " .. tostring(values))
```
burnhamrobertp: If an archive ships modes without this file, values stays nil and toVal errors rather than falling back, so Resolve throws straight out of Accept and out of the skirmish path. The fallback scan just below is there for archives that predate the modules layout, so isn't that pairing expected? UseValues also only runs on a modesByGame miss, so `values` ends up being whichever game loaded last rather than the one in hand.

LuaMenu/widgets/gui_battle_room_window.lua
```diff
@@ -64,6 +64,141 @@ local isMapDownloading = false
...
+				end
+			end
+		end
+	end
+
+	local defs = WG.ModoptionDefs
+	if defs then
+		for i = 1, #defs do
+			local opt = defs[i]
+			if opt.key and opt.section == category and opt.key ~= (category .. "_mode") then
```
burnhamrobertp: This matches opt.section against the category directly, but BAR puts mode_category on options_main and leaves section `game` holding only game_mode, which the condition excludes. So wouldn't this always come out empty? sectionCategory here and governedSections in moderesolver.lua both do the mode_category lookup.

LuaMenu/widgets/gui_modoptions_panel.lua
```diff
@@ -655,11 +832,331 @@ local function PopulateTab(options)
...
+			if catPolicy.allowRanked == false then
+				rankedPinned = true
+			end
+		end
+		isProgrammaticUpdate = true
+		if rankedPinned then
+			if rankedBeforeModePin == nil then
+				rankedBeforeModePin = localModoptions["ranked_game"] or modoptionDefaults["ranked_game"] or "1"
+			end
+			localModoptions["ranked_game"] = "0"
```
burnhamrobertp: mode_builder already covers this: .Ranked(false) adds a mode.ranked policy that puts ranked_game into modOptions, so the pin travels in the preset. On this path ranked_game isn't in mode.modOptions, so doesn't applyModeValues reset it to its default a few lines down, and isn't it in managedKeys anyway so nothing goes out? Nothing reaches it today since all four modes call .Ranked(), but the chain default is allowRanked = false, so a mode that just omits it would.

LuaMenu/widgets/gui_modoptions_panel.lua
```diff
@@ -655,11 +832,331 @@ local function PopulateTab(options)
...
+		pendingSeparator = nil
+
+		-- Options the mode pins outside the panel's sections (a mission pins
+		-- forceallunits, which lives with the cheats): rendered here so
+		-- picking the mode shows everything it does. Options hidden from the
+		-- lobby stay hidden — the pin still applies, silently.
+		if mode.modOptions and modoptions then
+			for i = 1, #modoptions do
+				local opt = modoptions[i]
+				if opt.key and mode.modOptions[opt.key] and not renderedKeys[opt.key]
```
burnhamrobertp, line 1066: ffa_wreckage sits in section `options`, which declares no mode_category, so it keeps its own tab and this renders it a second time in the mode panel. Both register modoptionControlNames[key] and the tabs loop is a pairs() walk, so SetControlLock looks like it only reaches whichever was built last, leaving the other copy editable with no lock on it. Since b69fc5b3 that edit can't reach the host either, so I think it just vanishes on Accept.

LuaMenu/widgets/gui_modoptions_panel.lua
```diff
@@ -1094,6 +1766,29 @@ local function InitializeModoptionsDisplay()
...
+		if activeModes and not applyingModes then
+			for cat in pairs(activeModes) do
+				local battleKey = panelModoptions[cat .. "_mode"]
+				if battleKey and selectedModeKeys[cat] ~= battleKey then
+					selectedModeKeys[cat] = battleKey
+					local ui = modeUI[cat]
+					if modoptionWindowOpen and ui and ui.applyMode then
+						local name = ui.itemKeyToName and ui.itemKeyToName[battleKey]
+						if name and ui.modeList then
+							ui.modeList:Select(name)
```
burnhamrobertp: ComboBox:Select with a string calls OnSelectName (combobox.lua:70), which queues another applyMode through WG.Delay, and then the line below calls applyMode directly. So a host mode change looks like it rebuilds the whole panel twice, 50ms apart, with the second pass landing after applyingModes is back to false.

LuaMenu/widgets/gui_battle_room_window.lua
```diff
@@ -4773,7 +4909,16 @@ local function InitializeControls(battleID, oldLobby, topPoportion, setupData)
 			local title = string.sub(message, string.find(message, ' "',nil,true) + 2, string.find(message, '" ', nil, true) - 1)
 			title = ShortenBSetCommand(title)
 			title = title:sub(1, 1):upper() .. title:sub(2)
-			votePanel.VoteUpdate(title,nil, ismapppoll, candidates, votesNeeded, mapname, userwhocalledvote, newlycalledvote)
+			local modeTooltip
+			local _, parsedModeKey = ParseModeVoteTitle(title)
+			if parsedModeKey then
+				title, modeTooltip = FormatModeVoteTitle(title)
+				Spring.Echo("[ModeVote] Display: " .. title)
```
burnhamrobertp: This branch also handles the countdown re-announcements (the `..s remaining.$` match further up), so these fire over and over for a single vote rather than once, and FormatModeVoteTitle gets re-run with them. Leftover debugging?

.gitignore
```diff
@@ -1,3 +1,9 @@
 devmode.txt
 *code-workspace
 .vscode/settings.json
+
+# Game mode presets (BAR export_game_modes widget). The ModeCommand plugin
+# self-fetches these from the BAR `game-modes` release into game_modes.cache.json;
+# game_modes.json is an optional host-dropped fallback. Neither is source.
+spads-plugins/ModeCommand/game_modes.json
```
burnhamrobertp: eeacba66 moved the plugin out to spads_config_bar, so these point at a directory this PR no longer adds, and they didn't match its filenames anyway (it read modes.json and wrote modes.cache.json). Worth dropping the hunk?

burnhamrobertp (changes requested):
Read this alongside BAR#8486 rather than on its own, so a lot of the below is me checking that the two halves line up; if I have the wrong BAR branch in mind then several of these fall away.

Separately, and not worth their own comments: moderesolver.lua still has no tests, though 05f5341a removes most of the reason to worry about that; the `libs/json.lua` include at gui_modoptions_panel.lua:2131 has no json use left anywhere in the file; IsOptionalModoption in gui_battle_room_window.lua:85 is never called; and the mode desc height at :997 guesses a 900px wrap against a TextBox sized with right = 15.

LuaMenu/widgets/gui_modoptions_panel.lua
keithharvey: right, it stored the key before applyMode compared it. Now it moves the selector without the listener and applies once.

LuaMenu/widgets/gui_modoptions_panel.lua
keithharvey: uses game_mode's def now.

LuaMenu/widgets/gui_modoptions_panel.lua
keithharvey: mode locks are their own bit; item locks keep their row where they put it.

LuaMenu/widgets/gui_modoptions_panel.lua
keithharvey: mode panel rows are stamped and the filter walks the scroll.

LuaMenu/widgets/gui_modoptions_panel.lua
keithharvey: per archive now. On a missing values.lua I chose to ignore the modes and log, rather than carry a copy for archives that predate the lib.

LuaMenu/widgets/gui_battle_room_window.lua
keithharvey: resolves through mode_category now.

LuaMenu/widgets/gui_modoptions_panel.lua
keithharvey: you're right and my earlier reply on this was stale: the preset carries the pin, the panel no longer touches ranked_game, and the builder pins it for a mode that never says Ranked() too.

LuaMenu/widgets/gui_modoptions_panel.lua
keithharvey: claimed options render only in the mode panel now.

LuaMenu/widgets/gui_modoptions_panel.lua
keithharvey: gone with the host-change fix above.

LuaMenu/widgets/gui_battle_room_window.lua
keithharvey: echoes and the dead helper removed.

.gitignore
keithharvey: dropped.

PR #1048, opened 2025-09-16
Update Chobby json.lua to match BAR json.lua

WilliamPas:
Should hopefully resolve an issue that prevents the latest map "Devil's Postpiles" from being downloaded.

### DIFF
--- a/libs/json.lua
+++ b/libs/json.lua
@@ -3,40 +3,45 @@
 -- json Module.
 -- Author: Craig Mason-Jones
 -- Homepage: http://json.luaforge.net/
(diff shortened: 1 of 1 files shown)


WilliamPas:
> "should"? Have you ran chobby with this change and tested that everything works?

Nope, I don't have that set up at all. Was just following your earlier suggestion.

p2004a:
Nice, thank you!

I've now just ran locally, I have the mipmap etc working so, not sure what is wrong with your checkout.

p2004a (approved):

PR #1050, opened 2025-09-17
Update json.lua, this time with proper changes.

WilliamPas:
Bringing over an updated version of json.lua, and re-writing small parts of public functions to maintain previous functionality.
Originally this was done to resolve an issue with one of the new maps failing to download, but caused other issues due to the decode function of the updated script to return null (for some reason?)

Well now the public decode function returns the internal decode function, and the public function names were changed back to their old names. Downloading works, as well as replays and option preset. I'm currently not aware of any other features that use json, so I would appreciate the advisement on such features so I can test them as well.


WilliamPas:
Replays working again.
image
Option presets working.
image
New map downloading successfully.
image

p2004a:
If it works, we *could* merge it to unblock map download, but I agree that changing the code to pollute the global scope is not optimal. Moreover, as I've pointed out in revert PR https://github.com/beyond-all-reason/BYAR-Chobby/pull/1049 The same bugged library for whatever terrible reason is in 2 other copies in the repository, we should not have multiple version in multiple copies.

Now, I don't mind having bad code in the repo *temporarily* to fix map download for players IF we will have a follow up to fix all properly. Otherwise it will be in half-assed state forever potentially causing more weird issues.

efrec:
I imagine this does load and run fine, but it has the wrong module pattern, still.

The module is still loaded repeatedly despite both acting as a dependency and being global. That's partly forgivable since it is the status quo.

You replaced the global `json` object with three global functions. There's no reason to have any globals in modules or libs in Lua, frankly.

I'd recommend this article again: https://blog.separateconcerns.com/2014-01-03-lua-module-policy.html

Which ends with the important bit: "I think this is what matters the most in the end: modules always return tables and never create globals. The rest is mostly implementation details!"

Issue #611, opened 2024-03-23
Buffer is not cleaned up on disconnect nor on establishing new connection


AntlerForce:
![Screenshot_20240315_164106](https://github.com/beyond-all-reason/BYAR-Chobby/assets/124457076/11f21375-2566-46b3-a787-aca5ec6b7557)
[spring-launcher-20231104T212641.log](https://github.com/beyond-all-reason/BYAR-Chobby/files/14733436/spring-launcher-20231104T212641.log)

"It looks like this IP is opening connection, getting the lobby server TASSERVER 0.38-33-ga5f3b28 * 8201 0 stuff, waiting for a few seconds, dropping connection, and doing like this in the loop"

https://github.com/beyond-all-reason/BYAR-Chobby/blob/3f044810841a32ac767e3306a125483e78fbe2b9/libs/liblobby/lobby/interface_shared.lua#L309

"This is bug in Chobby. Buffer is not cleaned up on disconnect nor on establishing new connection. So there can be garbage in it from previously closed connection."

![Screenshot_20240323_125709](https://github.com/beyond-all-reason/BYAR-Chobby/assets/124457076/7fbb585b-8a23-4b6f-9651-99252f14f2e3)

PR #1054, opened 2025-09-27
Add missing network timeout params

dzosz:
The purpose of this PR is to:

1. Enable reconnect feature which is used in clients to continue the game when network (address) changes
2. Increase the value of `NetworkTimeout` from 2 to 5 mins which is a deadline for communication with the server. Extended based on my personal experience on how long it takes me to setup alternative network connectivity.
3. Parameter `InitialNetworkTimeout` needs to be non-zero to prevent game from stopping when in pregame (before sim frame).

Was configuration `ReconnectTimeout` disabled with premeditation? Was it causing problems? I have tested it in pregame and after and it worked fine.


AntlerForce:
How can I test that this is working as intended?

dzosz:
Thanks. I haven't looked into lobby yet. I'll see if there's anything that we can do about it.
Few questions I will try to find an answer to:

1. Is lobby connection separate? Why is it failing to reconnect? Are the disconnects somehow related to spring engine reconnects or maybe they just use smaller timeouts that also can be updated?
2. Does the connection with lobby matter at all? I.e. will there by any negative side effects of disconnecting from lobby but still completing the game?

AntlerForce:
> 1. Is lobby connection separate? Why is it failing to reconnect? Are the disconnects somehow related to spring engine reconnects or maybe they just use smaller timeouts that also can be updated?

I believe this is due to the command buffer sending garbage or sending too many commands after reconnection and causing flood protection to kick the user. I've looked into this before but didn't have a good enough understanding to remediate it. There are more details about this issue here: https://github.com/beyond-all-reason/BYAR-Chobby/issues/611

AntlerForce:
@Beherith I searched each of these 3 settings in the BAR discord and noticed that you have changed around these values in the past, I would appreciate it if you could share your thoughts on changing these values if you have time.

PR #1056, opened 2025-09-30
Unofficial Maps

Jandodev:
Added to support potential unofficial map change allowing users to find and use unofficial maps in multiplayer.

https://github.com/beyond-all-reason/infrastructure/issues/8


p2004a:
(deleted my previous comment), ok, I get it, you want to make any map available locally selectable in multiplayer. I don't believe that selection of that should be based on what is available locally, and we do not want to allow people selecting anything that they have available locally.

Jandodev:
Understood, my only counter to this is wouldn't we want the players after they have loaded and downloaded another custom map to have the ability to search for it in the future?

Their probably is a much more elegant way of doing this like you said with the inf repo

p2004a:
Any local maps are available in single player so they have those available.

If we wanted to have a search function for custom maps, chobby would need to call some search endpoint or something, or we  just tell people to browse online.

Any chobby improvements can follow from basic working implementation.

Issue #1060, opened 2025-10-05
Outsmart the Barbarians missing rocks


ryanbennitt:
The Outsmart the Barbarians scenario mentions rocks that can be reclaimed, and old playthroughs on youtube show these rocks. However in the current version of the game these rocks are missing. In the infolog the following errors can be seen:

[t=00:00:55.084336][f=-000001] [LoadScreen::SetLoadMessage] text="Initializing Map Features"
[t=00:00:55.085202][f=-000001] Error: [LoadFeatureDefsFromMap] unknown map default feature-type "rock1" (only "treetype" and "geovent" are recognized)
[t=00:00:55.085245][f=-000001] Error: [GetFeatureDef] could not find FeatureDef "rock1"
[t=00:00:55.085300][f=-000001] Error: [GetFeatureDef] could not find FeatureDef "rock1"
[t=00:00:55.085310][f=-000001] Error: [GetFeatureDef] could not find FeatureDef "rock1"
[t=00:00:55.085328][f=-000001] Error: [GetFeatureDef] could not find FeatureDef "rock1"
[t=00:00:55.085338][f=-000001] Error: [GetFeatureDef] could not find FeatureDef "rock1"
[t=00:00:55.085348][f=-000001] Error: [GetFeatureDef] could not find FeatureDef "rock1"
[t=00:00:55.085357][f=-000001] Error: [GetFeatureDef] could not find FeatureDef "rock1"
[t=00:00:55.085366][f=-000001] Error: [GetFeatureDef] could not find FeatureDef "rock1"
[t=00:00:55.085375][f=-000001] Error: [GetFeatureDef] could not find FeatureDef "rock1"
[t=00:00:55.085388][f=-000001] Error: [GetFeatureDef] could not find FeatureDef "rock1"
[t=00:00:55.085418][f=-000001] Error: [GetFeatureDef] could not find FeatureDef "armshlt_dead"

Checking the contents of supreme_crossing_1 map, it contains definitions of rocks in the older TA style .tdf file, maybe this format isn't supported any more? [Discussion on discord](https://discord.com/channels/549281623154229250/1415094211224797335) suggests the map isn't going to be fixed.

But the scenario could be fixed, all it needs is some wreckage added, each valued around 500, to replace the rocks that have disappeared, e.g. the commander produces a heap valued at 500...

I'll raise a pull request to add this to the scenario shortly...

PR #1061, opened 2025-10-05
Issue #1060 Add wreckage valued at 500 to replace the missing rocks

ryanbennitt:
Added arm/corcom_heaps valued at 500 to replace the missing rocks that are broken in current versions of BAR.


AntlerForce:
Why commander wrecks instead of rocks?

ryanbennitt:
Because they're worth 500 metal to match the missing rocks, and they're not resurrectable.

Does supreme isthmus have rocks already? It's a bigger map though, isn't it?

PR #1071, opened 2025-10-12
Hide Newbie OS

Damgam:
Hides OS of players above 6 sigma (uncertainty)
Sorts players in a way that the unknown ones are at the bottom.
obraz
obraz

BAR side PR: https://github.com/beyond-all-reason/Beyond-All-Reason/pull/5925


NortySpock:
While we're here, does chobby support displaying negative OS? (How easy is that to test?)

I recall an issue in teiserver asking to send negative OS values to the client,  but there were questions around compatibility.

PR #1081, opened 2025-11-12
Add links (for linking feedback threads/forms) to modoptions panel.

robertthepie:
### Work Done
Adds a link type of modoption so that links to surveys can be added next to modoptions

these links open a in game pop up first to confirm that the user wants to open the external link

The tooltip is just the link for link modoption entries
The desc is still used, but instead of tooltip it can be used to describe the link in the confirm open link popup

Also, if hidden is a number, checks it against current unix time to determine if it should be hidden so that these survey links can be hidden after some time, by extension allowing other modoptions to be hidden after some time.

### Setup
To see this in action, set your chobby to this pr, and as well as chobby to use beyond all reason dev.
Add the following entry into your modoptions file before opening skirmish
```lua
    {
        -- since this isn't a real modoption they should be fine to all share it
        key     = "link",
        -- button label
        name    = "View list of cheats Here",
        -- desc instead of tooltip is used on the confirm open screen, at aproxcimatly 34 characters it overflows to next line,
        -- you have roughly 4 lines to work between this and the link, so round up link length div by 34, sub from 4, how many lines you can have
        desc    = "[Keep this very breef, ~34 chars.]",
        section = "options_cheats",
        -- new modoption panel type
        type    = "link",
        -- the link to open
        link    = "https://discord.com/channels/549281623154229250/549282587487502347/1438232160539639818",
        -- use this to make the button not look silly, adjust based on name length
        width   = 250,
        -- use the column to put it next to the button it is meant to accompany
        -- anything more than 1 will put it to the right, 2 is just next to the modoption, bot for some it might be better to move it left or right
        column  = 2,
    },
```

### Image Example:
image

### DIFF
--- a/LuaMenu/images/License.txt
+++ b/LuaMenu/images/License.txt
@@ -21,3 +21,5 @@ spectating.png
 startboxsplit*.png (all)
 unknown_sync.png
 unready.png
+
+The file external-website.png is "External Link Icon" by Raj Dev, sourced from FreeIcons (https://freeicons.io/icon/business-and-online-icons/external-link-icon-icon), and is licensed under Creative Commons Attribution 3.0 (https://creativecommons.org/licenses/by/3.0/). Modified from the original.


LuaMenu/widgets/gui_modoptions_panel.lua
```diff
@@ -1065,6 +1105,9 @@ function ModoptionsPanel.LoadModoptions(gameName, newBattleLobby)
 				modoptionDefaults[data.key] = tostring(data.def)
 			end
 		end
+		if data.hidden and type(data.hidden) == "number" then
+			data.hidden = data.hidden < unixDate
```
NortySpock: Why hide the link from the user?

Shouldn't `hidden` be a Boolean rather than a time?

robertthepie: I think the confusion would more-of lie game side, where hidden would now ocasionally be a number,
but i don't want it to be a new field to avoid a: why is it hidden when hidden is false even if there is a hide on day option, as literacy is a rare skills to have

SethDGamre:
If it's serviceable, send it says I
The thing about hidden Boolean/number sounds like some niche nerd stuff that shouldn't hold up a monumental step forward in collecting feedback from non-discord addicts.

PR #1118, opened 2026-01-19
Add BarbarianAI difficulty profile to displayName

Arcadiyyyyyyyy:
Displays the profile name in the lobby if specified, to decrease the confusion of the skirmish setup 

Result visualisation: 
https://github.com/user-attachments/assets/c0ed7ec7-d5a2-45b2-bb53-37d7f7e7cc2e

Does not display the default profile values, as the game engine selects them as far as I could tell


AntlerForce:
This looks pretty good, I just have a couple of comments:

> Does not display the default profile values, as the game engine selects them as far as I could tell

Default profile is Barb spawning as hard-balanced, we can assume this and label accordingly.

Adding profile to display name isn't working with Simple Skirmish, you can test this by clicking Advanced at the end of the Simple Skirmish setup instead of Start. I don't think this is strictly needed but would be nice to have.

PR #1143, opened 2026-03-02
barbai profile remembers previous selection

ZainMGit:
When adding a BARbarian AI without selecting a profile using options,  it uses the ``lastBarbAiProfile`` which was selected.  Default for first time selection is the Easy difficulty profile.

### DIFF
--- a/LuaMenu/widgets/chobby/components/ai_list_window.lua
+++ b/LuaMenu/widgets/chobby/components/ai_list_window.lua
@@ -198,19 +198,17 @@ function AiListWindow:AddAi(displayName, shortName, version, options)
 		end
 		counter = counter + 1
 	end
-	
+
 	local battleStatusOptions = {side = math.random(0,1), teamColor = PickRandomColor(),}
 
 	if shortName == "BARb" then
 		if type(options) == "table" and options.profile and options.profile ~= "" then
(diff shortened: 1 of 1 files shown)


kroIya:
Why change the default?

AntlerForce:
> Why change the default?

if a default setting has to mutually exclusively cater to new players or non-new, it should cater to new players because they have a zero chance of already having configured settings to their liking. Players can choose higher difficulties after defeating the easier one, and now it's remembered.

Easy default BARB AI is a logical choice that I suggested and @ZainMGit followed through on.

I tested and didn't find any issues, and it's very nice to have my AI difficulty choice remembered after restart. I initially wondered if having separate AI difficulty settings in the lobby settings tab and AI spawn menu would be confusing, but I think I like being able to try new AI difficulties and get better without "polluting" the random config.

kroIya:
@AntlerForce 
If you're backing the change, then I suppose so shall it be.
I'd urge you to reconsider though, Easy barb is not only easier, it's purposefully anemic and doesn't showcase the AI in the best light. It's also not obvious that difficulty settings exist, and a player who'd look to move up from the easy ai is not yet the kind of experienced veteran who's comfortable changing the settings.

PR #1152, opened 2026-03-08
Auto rejoin 

### LINKED ISSUE #1150: Auto rejoin should not attempt to reconnect to battles older than 1-2 hours or so
Apparently the rejoin prompt will suggest games from the previous day if a crash occurred, it probably should not bother if the match was from long enough ago that it can be reasonably expected to be over.

ZainMGit:
Fixes #1150 

Auto rejoin only prompts if it has been <4 hours since battle join.

Creates ``rejoinBattleTimestamp `` at the same time which ``rejoinBattleID`` is created. if the timestamp is >4 hours then the rejoin prompt does not display. 

4 hours is probably the maximum I'd see a match going but value can be adjusted.


LuaMenu/widgets/gui_battle_rejoin.lua
```diff
@@ -15,17 +15,32 @@ end
...
 
 local function LoginRejoinOption(listener, battleID)
 	-- Check if a battle ID has been stored for rejoining.
 	if not WG.Chobby.Configuration.rejoinBattleID then
 		return
 	end
+
+	local rejoinBattleTimestamp = tonumber(WG.Chobby.Configuration.rejoinBattleTimestamp)
+	if not rejoinBattleTimestamp then
+		WG.Chobby.Configuration:SetConfigValue("rejoinBattleID", nil)
```
AntlerForce: We should clear `rejoinBattleTimestamp` here as well, just in case it can't be parsed because it is malformed rather than just nil.

AntlerForce: Actually `tonumber` will return nil on failure, not false, so on second thought I think this is okay.

RogShotz:
Oh hey I commented on 1150. I think that reminding the player that they can reconnect from a crash is a better solution than no display. I kind of half assed my rejoin button to remind them that they just full left the game after the crash, but, after 4 hours changing the text to be "You previously crashed, remember that you can always rejoin!" might be better. Just my thoughts :)

It also might be easier to implement js saying ;)

PR #1153, opened 2026-03-09
Add Support for Experimental King of the Hill Game Mode

TTNO1:
See https://github.com/beyond-all-reason/Beyond-All-Reason/pull/6894

This PR adds a hill area selection box, identical to the start box selection, for the PR linked above.

image


AntlerForce:
I'm not exactly sure what I'm looking at here but boxes that are expected to be defined through arbitrary modoptions every time instead of set via rowy and passed via maps-metadata is not appealing to me in the slightest.

PR #1158, opened 2026-03-15
add open replay folder button to replay panel

Varunda:
this PR adds a button to the replays panel in lobby that opens the folder with all the replays (specifically the demos folder within the `writePath`)

this button was placed to the left of the `Refresh` button

the current quickest way to open the replays folder (when the game is opened) is going to `Links > Game Data`, then opening the demos folder from there. this adds a button to just open the folder with the replays

also changes the `x` position of the buttons to be based on multiples of 130, which doesn't move them at all, and uses i18n for the text and tooltip of the new button

only tested on Windows

no AI was used

## screenshots

screenshot with devmode enabled:
image

without devmode:
image

### DIFF
--- a/LuaMenu/widgets/gui_replay_handler.lua
+++ b/LuaMenu/widgets/gui_replay_handler.lua
@@ -725,13 +725,13 @@ local function InitializeControls(parentControl)
 		}
 
 		local manualbutton = Button:New{
-			right = 230,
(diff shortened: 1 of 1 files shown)


AntlerForce:
> image
> easier for me to just throw the debugMode options onto a new line, that solution fine?

Sure.

PR #1161, opened 2026-03-22
Widgets manager

WybrenKoelmans:
Download, install and upgrade widgets from CDN

Also see:
https://github.com/WybrenKoelmans/Public-BAR-Widgets

AI:  mostly working pseudo code into actual chobby code, adding error handling, documentation, used variety of models based on the task at hand

TODO:
- [x] reset and repopulate widgets source repo with list of initial (curated) widgets
- [x] transfer widgets source repo to BAR org
- [x] move CDN source to BAR owned


LuaMenu/widgets/chobby/components/plugins_window.lua
```diff
@@ -0,0 +1,1358 @@
...
+                y = 150,
+                right = 8,
+                height = 40,
+                fontSize = 14,
+                autosize = false,
+                wordwrap = true,
+            },
+            -- Install button
+            Button:New {
+                caption = isWidgetInstalled(id) and "Installed" or (installingWidgets[id] and "Installing..." or "Install"),
```
PtaQQ: You added plugins = "Plugins" to i18n but UI strings bypass i18n().

LuaMenu/widgets/api_download_handler.lua
```diff
@@ -273,9 +316,12 @@ function externalFunctions.QueueDownload(name, fileType, priority, retryCount, r
 		id = downloadCount,
 		retryCount = retryCount or 0,
 		resource = resource,
+		hidden = (resource and resource.hidden) and true or false,
 	}
 	requestUpdate = true
-	CallListeners("DownloadQueued", downloadCount, name, fileType, resource)
+	if not downloadQueue[#downloadQueue].hidden then
```
PtaQQ: Prefer checking resource and resource.hidden directly; the dedupe early-return in QueueDownload can desync this.

LuaMenu/widgets/chobby/components/plugins_window.lua
```diff
@@ -0,0 +1,1358 @@
...
+end
+
+--------------------------------------------------------------------------------
+-- Initialization
+--------------------------------------------------------------------------------
+
+function PluginsWindow:init(parent)
+    -- Register download event listeners
+    if WG.DownloadHandler and WG.DownloadHandler.AddListener then
+        WG.DownloadHandler.AddListener("DownloadFinished", onDownloadFinished)
```
PtaQQ, line 1296: Closures keep firing after the window is disposed.

LuaMenu/widgets/chobby/components/plugins_window.lua
```diff
@@ -0,0 +1,1358 @@
...
+    if ok then
+        Spring.Echo("[PluginsWindow] Backed up directory: " .. cleanPath .. " -> " .. backupPath)
+        -- Rename all .lua files to .lua.backup so the engine won't load them
+        local luaFiles = VFS.DirList(backupPath .. "/", "*.lua")
+        if luaFiles then
+            for _, luaFile in ipairs(luaFiles) do
+                os.rename(luaFile, luaFile .. ".backup")
+            end
+        end
+        local subDirs = VFS.SubDirs(backupPath .. "/")
```
PtaQQ: Deeper *.lua under _backup_ stay active. Recurse, or move backups outside Widgets/.

LuaMenu/widgets/chobby/components/plugins_window.lua
```diff
@@ -0,0 +1,1358 @@
...
+            if content then
+                parseManifest(content)
+            else
+                loadState = STATE_ERROR
+                loadError = "Could not read downloaded manifest file"
+                Spring.Echo("[PluginsWindow] Could not open manifest file at: " .. MANIFEST_DEST)
+            end
+        end
+        -- Auto-upgrade installed widgets that have newer versions available
+        if loadState == STATE_LOADED then
```
PtaQQ, line 1075: should be opt in, not automatic (with no user action)

LuaMenu/widgets/chobby/components/plugins_window.lua
```diff
@@ -0,0 +1,1358 @@
...
+-- Ensure json module is available
+local json = json or VFS.Include and VFS.Include("libs/json.lua") or nil
+
+PluginsWindow = LCS.class{}
+
+--------------------------------------------------------------------------------
+-- Constants
+--------------------------------------------------------------------------------
+
+local CDN_BASE         = "https://bar-workshop.zen-ben.com/bar-workshop"
```
PtaQQ: highlighting this just to make sure it gets relinked

LuaMenu/widgets/chobby/components/plugins_window.lua
```diff
@@ -0,0 +1,1358 @@
...
+
+local function checkForUpgrades()
+    if #widgetsList == 0 then return end
+    for _, widget in ipairs(widgetsList) do
+        local widgetId = widget.id
+        if widgetId and widget.version and isWidgetInstalled(widgetId) then
+            local installedVersion = getInstalledVersion(widgetId)
+            if compareVersions(installedVersion, widget.version) < 0 then
+                Spring.Echo("[PluginsWindow] Upgrading " .. widgetId .. " from v" .. tostring(installedVersion) .. " to v" .. tostring(widget.version))
+                if installingWidgets[widgetId] then
```
PtaQQ: Should skip this iteration only (goto continue / restructure), not bail out.

LuaMenu/widgets/chobby/components/plugins_window.lua
```diff
@@ -0,0 +1,1358 @@
...
+    do
+        local img = Image:New {
+            x = rightX,
+            y = contentY,
+            right = 10,
+            height = "40%",
+            keepAspect = true,
+            checkFileExists = true,
+            file = coverPath,
+            fallbackFile = IMG_FALLBACK,
```
PtaQQ: Use IMG_FALLBACK_MEDIUM for cards, IMG_FALLBACK_LARGE for the cover.

PtaQQ (changes requested):

PtaQQ:
One more thing, could you add a popup to show after clicking install with "I understand BAR organisation is not responsible for this custom code." with a "don't show again" checkbox?

LuaMenu/widgets/chobby/components/plugins_window.lua
WybrenKoelmans: Although currently the game does not load lua deeper than one sublevel, it doesn't hurt to prepare for when it does.

LuaMenu/widgets/chobby/components/plugins_window.lua
WybrenKoelmans: Could you check with Marek for the link (and setting up the correct GitHub secrets for the CI)?

LuaMenu/widgets/chobby/components/plugins_window.lua
WybrenKoelmans: If we are already upgrading this widget, it means that all widgets are currently being updated, so this iterator can bail out.

LuaMenu/widgets/chobby/components/plugins_window.lua
WybrenKoelmans: I'd personally be really annoyed having to manually keep my list of widgets up to date. We'd also need some type of notification mechanism in case we do want this. What is the motivation for not doing it automatically? We generally need to protect users from running very old versions of widgets that are suboptimal and much improved in later versions IMO.

WybrenKoelmans:
image
image

OK! 
[x] I added  two buttons for the Source/Discord links. 
[x] I fully reset and refilled and rechecked all widgets in the WidgetHub repo. 

We'll now have 40 widgets currently, I tried to fill the first page with the most popular/functional widgets from the Discord channel, after that my widget repo is appended and I took out the non GDD conform ones.

So as far as i'm concerned we are good to go, do you guys mind giving it another test run?

LuaMenu/widgets/chobby/components/plugins_window.lua
AntlerForce: I see pros and cons to either choice, at the moment I am unsure which is best. Automatic updating gives better ability to widget developers to fix broken widgets without needing user intervention. A non-zero amount of my time in the support channel has been spent looking into issues only to find undisclosed user-installed broken out of date widgets.

On the other hand this also enables widget developers to break user interfaces if not caught before distribution, and create confusion as to what happened. No version control or uninstallation in plugin interface means that if this happens the only solution is manual intervention or waiting for a fix to roll out, as there is no pathway to rolling back.

AntlerForce:
Screenshot_20240869

This keeps coming up for me after a restart even if I check do not ask again, this key needs to be saved to `configuration.lua`. Example: https://github.com/WybrenKoelmans/BYAR-Chobby/blob/6fceec639c5b8d69fe5560db436d964c20ce827b/LuaMenu/widgets/chobby/components/configuration.lua#L222

LuaMenu/widgets/chobby/components/plugins_window.lua
AntlerForce: We have settled on automatic updates by default with notifications for any install/update, and a toggle to opt-out under lobby settings.

PR #1179, opened 2026-04-13
add Nation Wars 2026 registration announcement

nbusseneau:


nbusseneau:
2026-04-13_02-58-33_spring

LuaMenu/configs/gameConfig/byar/welcomePanelItems.lua
```diff
@@ -34,27 +42,6 @@ local welcomePanelItems = {
         UrlText = "Read the Announcement",
     },
 
-    { -- Please don't delete this, just comment it out, it's getting reused.
```
AntlerForce, line 37: May as well comment this whole entry out like this comment requests.

nbusseneau: I saw it but I think this is bad practice, hence why I removed it instead -- commenting entries that may or may not be re-used someday is just adding bloat for no reason, since it makes absolutely no difference to add it back from the last commit that added it.

nbusseneau: That said, if all other contributors think this is OK I can edit it back in -- I just really think it's bad practice to bloat files with comments but community > personal preference 😅

AntlerForce: @Damgam Can you archive this elsewhere?

nbusseneau: It's already archived in this very repo => `git cherry-pick 633e5d212a692826b1af26b459ae824239257f1f` 😁

PR #1182, opened 2026-04-22
Delete LuaMenu/images/flags/sy.png

Snipebusher:
Will be replaced with new flag
# Options:
## Simply downscaled:
Flag_of_Syria_(2025-) svg

## Downscaled and edited:
Flag_of_Syria_(2025-) svg(1)


ChocoboCoding:
Maybe I am missing something, but shouldn't you replace the flag in the PR, instead of just deleting it?

Snipebusher:
I should, but I want to know which of the two flags I should use.

PR #1192, opened 2026-05-02
fix(presets): can load very large presets via batching (20k chars every 4 seconds)

tetrisface:
95% done with AI.

# Throttled Preset Loading

## Context

Large saved option presets can emit enough battle commands to hit Teiserver's Spring command rate limit. The observed failure mode is a user being kicked while applying a large preset. The backend rate limit exists to protect server memory, so the client-side fix is to avoid bursts rather than bypass the limit.

This change keeps the existing saved preset flow, but sends modoption changes in delayed batches.

## Changed Files

- `libs/liblobby/lobby/interface.lua`
  - Adds `Interface:SetModOptionsThrottled(...)`.
- `LuaMenu/widgets/gui_optionpresets_panel.lua`
  - Routes saved preset modoption loading through the throttled sender.
  - Adds progress and cancellation UI for delayed loads.

## Throttled Sender

Signature:

```lua
Interface:SetModOptionsThrottled(data, maxCharsPerBatch, intervalSeconds, progressCallback, cancelToken)
```

Parameters:

- `data`: modoption key/value table.
- `maxCharsPerBatch`: maximum accumulated command text per batch. Defaults to `20000`.
- `intervalSeconds`: delay between batches. Defaults to `4`.
- `progressCallback`: optional callback as `function(current, total, cancelled)`.
- `cancelToken`: optional table; setting `cancelToken.cancelled = true` stops the next unsent batch.

The implementation mirrors `SetModOptions()` by skipping values that already match `self.modoptions`. Remaining commands are grouped by approximate command text size, then sent batch by batch with `WG.Delay()`.

Cancellation is cooperative. A batch already sent cannot be undone, but closing/cancelling the UI prevents later delayed batches from being sent.

## UI Behavior

Saved presets now show a small progress dialog while modoptions are being sent. The dialog updates the batch counter and progress bar after each batch and includes a cancel button.

Progress labels use Chili `SetCaption()` instead of mutating `caption` directly, because direct mutation does not reliably redraw labels.

The attempted raw text import path was intentionally removed. Large preset text in Chili edit boxes and `Spring.GetClipboard()` both proved capable of blocking the UI. The supported path for this change is saved preset loading.

## Rate Limit Choice

Current UI call sites use:

```lua
battleLobby:SetModOptionsThrottled(modoptions, 20000, 4, progressCallback, cancelToken)
```

This is intentionally conservative:

- Teiserver Spring command limit: roughly `200` commands/minute by default.
- Client sends at most one batch every `4` seconds.
- The `20000` character limit stays well below the default `64KB` message buffer.

The batching limit is character-based rather than count-based because tweakdef values can vary significantly in size. This keeps large values from forming very large single sends while avoiding protocol changes.

## Review Notes

- Existing `SetModOptions()` is unchanged.
- Non-modoption preset parts still apply immediately as before.
- The new throttling only affects saved preset modoption application.
- `WG.Delay()` keeps the UI responsive between batches.
- No retry or acknowledgement layer was added; that would require stronger server/protocol feedback than this client path currently has.

## Suggested Verification

- Load a saved preset with enough changed modoptions to produce multiple batches; confirm the batch counter and progress bar advance together.
- Cancel after the first batch; confirm no further delayed batches are sent.
- Close the preset window during a load; confirm pending batches are cancelled.
- Test a small/no-op preset; confirm the UI closes cleanly and no unnecessary commands are sent.

## References

- **Backend rate limit**: `teiserver.Spring rate limit per minute` (default: 200)
- **Buffer size**: `teiserver.Spring max message buffer size` (default: 64KB)
- **Teiserver Config**: https://github.com/beyond-all-reason/teiserver/blob/master/lib/teiserver/libs/teiserver_configs.ex

### DIFF
--- a/LuaMenu/widgets/gui_optionpresets_panel.lua
+++ b/LuaMenu/widgets/gui_optionpresets_panel.lua
@@ -243,6 +243,7 @@ local function createProgressDialog(parentWindow, title, onCancel)
 		right = 15,
 		height = 30,
 		align = "left",
+		autosize = false,
 		caption = "Loading settings...",
 		objectOverrideFont = WG.Chobby.Configuration:GetFont(2),
 	}
@@ -264,6 +265,7 @@ local function createProgressDialog(parentWindow, title, onCancel)
 		right = 15,
 		height = 30,
 		align = "center",
+		autosize = false,
 		caption = "0 / 0 batches",
 		objectOverrideFont = WG.Chobby.Configuration:GetFont(2),
 	}
--- a/LuaMenu/widgets/gui_optionpresets_panel.lua
+++ b/LuaMenu/widgets/gui_optionpresets_panel.lua
@@ -43,13 +43,13 @@ local currentMPBattleSettings
 local multiplayerModoptions
 
 -- now we need to store the object in this class
-local jsondata;
+local jsondata
 
 -- preset that is selected in the dropdown menu
-local selectedPresetName = placeHolder;
+local selectedPresetName = placeHolder
 
 -- preset that is currently applied
-local appliedPresetName  = placeHolder;
+local appliedPresetName  = placeHolder
 
 -- active throttled preset load, used to cancel delayed batches when closing UI
 local activeLoadToken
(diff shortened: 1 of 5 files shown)


tetrisface:
here is the largest test preset i know https://gist.github.com/tetrisface/400b71713248b3155e63f6ea13a1fd0c

LuaMenu/widgets/gui_optionpresets_panel.lua
```diff
@@ -209,11 +212,135 @@ local function applyPreset(presetName)
...
+		y = 15,
+		right = 15,
+		height = 30,
+		align = "left",
+		autosize = false,
+		caption = "Loading settings...",
+		objectOverrideFont = WG.Chobby.Configuration:GetFont(2),
+	}
+
+	-- Progress bar
```
AntlerForce: Please remove all useless comments like this. If in doubt about usefulness comments should explain why, not what. I don't need you to explain to me that `local progressBar` is a Progress bar.

AntlerForce (changes requested):

AntlerForce:
> here is the largest test preset i know https://gist.github.com/tetrisface/400b71713248b3155e63f6ea13a1fd0c

I need clearer testing steps please. Is it expected that I shouldn't be able to apply these options with Alpha but that it will work with your branch? Because although there is some lag applying these options seems to be working for me in a multiplayer host in Alpha when bossed, at least I am not triggering flood protection.

Can this be simplified down to a shared `optionsPresets.json` that contains a preset that breaks with Alpha that works with your changes?

tetrisface:
thanks for checking @AntlerForce 

* removed the three noisy comments. 
* reduced default interval from 5 to 4 seconds. 2 seems to work so 5 seems a bit much. 3 maybe too close for future changes.
* progress bar not shown now for single batch presets
* rearranged the progress window text to just one line
* removed the progress window header. i dont think chobby has that anywhere else.
* centralized magic numbers to one place

here is a gist with a 245k preset and a 9k preset optionsPresets.json file: 
https://gist.github.com/tetrisface/197fa6642141fd18006d3ab65b57c711

it should be possible to switch your regular one with this one for testing. 

in Alpha:
* with the 245k preset the behaviour is that the first time it simply fails to fully apply and only applies a partial setup. the second time i get force logged out.


in this branch:
* with the 245k preset it will show progress bar and start to batch.
* with the 9k it will just apply immediately without showing the progress window since its just one batch. so this is unchanged from current behaviour.

AntlerForce:
Thanks, I tried out the 245k preset but the room unlocked and many people flooded in, so I was able to get a test game with a replay and log.
https://www.beyondallreason.info/replays?gameId=62e1f86988c4dbfde497a519048c1a2a

[infolog.txt](https://github.com/user-attachments/files/27372968/infolog.txt)

There are some tweak related errors in the log but people said things seemed to be working fine ingame. And I never triggered flood protection or crashed the room.

Screenshot_20240792

I noticed the progress dialog was offcenter, also having multiple Cancel buttons clickable at this time might be confusing.

tetrisface:
yeah that tweak is quite old and unmantained. its just the largest tweak i have ever encountered. most mods dont really cross 100-120k i think. its basically only nuttyb variations that go up that far.

AntlerForce:
I figured the errors may be the tweak itself and not anything to do with the loading, thanks for confirming. If the above interface can be improved a bit then I am okay to merge this.

tetrisface:
I think I will be able to solve chat pasting throttling as well. Its a bit of a restructure. would you prefer me to bake that into this PR or a second one? it might not be done tomorrow. @AntlerForce

AntlerForce:
> I think I will be able to solve chat pasting throttling as well. Its a bit of a restructure. would you prefer me to bake that into this PR or a second one? it might not be done tomorrow. @AntlerForce

Either is fine, up to you.

tetrisface:
I will bake it into this PR. 

What do you think about truncating the chat output of large tweaks into something like "ABCDEF1234...ABCDEF1234" so that you get the beginning and ending? as well as printing previous modoption values so that you know what the value was before...?
I can add that as well.

AntlerForce:
> What do you think about truncating the chat output of large tweaks into something like "ABCDEF1234...ABCDEF1234" so that you get the beginning and ending? as well as printing previous modoption values so that you know what the value was before...? I can add that as well.

That sounds good, current interface is a bit ugly as large pastes in chat expand out of the box. In the past I looked into making the editbox expand to multiline with large pastes, but multiline editboxes were not functioning in chobby.

tetrisface:
done with testing now

tetrisface:
fixed the off-center issue. but now i saw the double cancel comment too. im thinking maybe its better to inline the progress in the parent window instead. with its own cancel button.
i think its fine as it is but i can fix it. will look later tonight.

AntlerForce:
Screenshot_20240797

The text on the progressbar is spilling over out of the bar on both sides.

tetrisface:
I think it's solved now. I tested single and double pane on 1920x1080 and 2560x1440.

AntlerForce:
LGTM, Thanks again.

tetrisface:
Great! Thanks! 🎉

PR #1213, opened 2026-07-06
Move online playercount

ZainMGit:
Previously proposed change, I created a new PR request to sort merge conflicts. Change not approved because the player count appeared while offline.

Moved online count to a more visible area. Value updates every 60s

image
image

Used codex to format the playercount text


AntlerForce:
Screenshot_20240850
Screenshot_20240849

I think font 11 looks a bit better than font 2 for onlineCountText.

Have you tested how this change accommodates max username length? I tried to do so myself but renaming is not working reliably for me. ~There seems to be some confusion over what the enforced max length is.~ 
I was able to rename to (Wx16) once and captured this screenshot, but after I renamed back and tried to rename to Wx16 again to see how it looked on alpha I can't do it anymore.
EDIT: 20 characters is max, renaming is capped at twice a week and thrice a month.
Screenshot_20240851

AntlerForce:
Screenshot_20240853

This is what a username with 19 Ws looks like on Alpha. 

Likely need to increase width + shift userControl x to the left to have enough room for everything.

Feel free to push changes + ping me to test instead of wasting a rename.

ZainMGit:
I can test using local variables, I don't need to use a rename. 
I moved it very far to the left to allow all 20 Ws. It can be moved more to the right and then truncated at 15 characters or so. 



image


image

AntlerForce:
Probably would look better after all if moved to the right and truncated. 

Good enough regardless, I'll probably do some tweaks on my own time.

Thanks!

PR #1217, opened 2026-07-07
Improve rename account UX

### LINKED ISSUE #1214: Renaming has terrible ux
The message returned upon pressing the change username button clips into the horizontal bar below. 

The box is not long enough to preview a name of significant length.

There is no report of success or failure even though this message is apparently sent by server (quote "something like SERVERMSG Max length 20 characters"). An example failure like Username being already taken is never communicated.

Max allowed length is never stated anywhere.

The existence of a rename cooldown (quote "no more than twice a week, 3 times a month") is never stated anywhere.

ZainMGit:
Changes Recover/Change UI in account settings"

Fixes #1214 

Provides error message if username is over 20 characters. States maximum allowed character length as well as stating maximum username changing frequency. 

Text no longer clips. 

image

used codex for formatting

### DIFF
--- a/LuaMenu/widgets/chobby/components/login_window.lua
+++ b/LuaMenu/widgets/chobby/components/login_window.lua
@@ -557,9 +557,9 @@ function LoginWindow:init(failFunction, cancelText, windowClassname, params)
 		x = pad + formw * 0 ,
 		y = pad + formh * 0 ,
 		width =   formw * 3 ,
-		height =  formh * 2 ,
+		height =  60 ,
 		-- caption = i18n("register_long"),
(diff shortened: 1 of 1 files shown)


AntlerForce:
Image

Nice, but this panel is now no longer handling rescaling gracefully. If I shrink the window a little scrollbars appear on the right and bottom, and the bottom of the buttons get cut off:

Image

On a 1366x768 screen with a taskbar the top and the bottom of the panel is cutoff:

Image

ZainMGit:
image

Properly scales on  1366x768 screen now.

AntlerForce:
Thanks!

PR #1226, opened 2026-07-13
asterix after username

ZainMGit:
Adds an ``*`` at the end of a player username if GetPlayerNote() returns true.  An asterix is not a character currently allowed to be used in a player username so it will not get confused with just being apart of the username. 

Appears across all places the username is recieved using ``aapi_user_handler``

upon adding or removing a playernote the playername is refreshed immediately. 

image

### DIFF
--- a/LuaMenu/widgets/api_user_handler.lua
+++ b/LuaMenu/widgets/api_user_handler.lua
@@ -766,6 +766,25 @@ local function UpdateUserStatusImage(userName, userControls)
 	end
 end
 
+local function UpdateUserDisplayName(userName, userControls)
+	if not userControls.tbName then
+		return
+	end
+
+	local displayName = GetPlayerDisplayName(userName, userControls)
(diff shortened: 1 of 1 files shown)


AntlerForce:
"upon adding or removing a playernote the playername is refreshed immediately."

This didn't seem to work for me, testing on myself and a friend in the friends list, asterix was not added or removed until I manually moved myself from spectator to player, or restarted lobby to reload friends panel.
Screenshot_20240870
Screenshot_20240872
Screenshot_20240871

ZainMGit:
Oh, it is because those names are not continuously updated with ``api_user_handler``.


Adding or removing a playernote  now refreshes tbName so the cached username is updated

AntlerForce:
Screenshot_20240874

Now the asterisk can appear in player status up top, can this be prevented? Or just block making notes on yourself like I have been.

ZainMGit:
Yes, it seems a bit silly to have a note for yourself. I can disable this.

PR #1233, opened 2026-07-21
Reset Widgets to Default Button

ZainMGit:
PR adds a ``Reset to Default`` button in the Widgethub. This button removes all custom widgets which a user has downloaded and restores in game widgets back to their default. 

Button requires a confirmation along with a disclaimer that all custom widgets will have to be redownloaded and that this action is final. 

After button is pressed:
 - ``LuaUI/widgets`` is cleared which removes Widgethub installed widgets and manually copied custom widgets from the widget folder.

- removes ``LuaUI/Config/IGL_order.lua``, ``IGL_known.lua``, ``IGL_data.lua``
       - this makes LuaUI rebuild these configs from the base game defaults
- Clears install cache from the widgethub and makes the widgehub recheck installed widgets so previously installed widgets show ``install`` instead of ``installed`` 

image
image

Used Codex GPT 5.5 to assist with programming


WybrenKoelmans:
> LuaUI/widgets is cleared which removes Widgethub installed widgets and manually copied custom widgets from the widget folder.

We can't assume that it's fine to just nuke a folder like that. In my case, you'd delete my whole git repository and history for my work on custom widgets. Other users might have symlinks to other parts of their system that will end up being recursively deleted. 

Take inspiration from this and why I think it's a bad idea: https://www.reddit.com/r/RotMG/comments/g1xyj0/rotmg_exalt_problems_with_installer_please_wait/?utm_source=share&utm_medium=ios_app&utm_name=iossmf

https://github.com/ValveSoftware/steam-for-linux/issues/3671

Even if we did only clear the actual widget directories, some widgets currently are modified by the users, a lot of time spend on configuring, etc, i'm not sure it's worth reclaiming ~50MB (probably much much less) for that.

You could trigger the backup mechnism that is currently in place for each widget, but not automatically download the latest version. That would be a much safer option.

Probably a way to reset all the configuration for custom widgets is good! Maybe even a "Full reset"  for not only custom widgets, but the normal options as well.

ZainMGit:
I wanted this button to act as a full defaults button not just for removing custom widgets but also changing any configured base game widgets back to normal. I think there should be a reset somewhere for this which does not require manually deleting folders in file explorer or reinstalling the game. 



Instead of removing items in the folder, would moving them to a "backups" folder be fine behavior?

ZainMGit:
Okay so I changed how the default button works. 

Reset to Default no longer deletes `LuaUI/Widgets` instead, the existing `LuaUI/Widgets` directory is moved to `LuaUI/WidgetBackups/Widgets_backup_<timestamp>`. Then a new empty `LuaUI/Widgets` directory is created.

This removes all custom widgets from being used in game and resetting the game to its default "state"  while preserving the original files.

LuaMenu/widgets/chobby/components/plugins_window.lua
```diff
@@ -33,6 +33,15 @@ local MANIFEST_URL     = CDN_BASE .. "/manifests.json"
 local MANIFEST_DEST    = "LuaUI/Widgets/manifests.json"
 local MANIFEST_NAME    = "plugin_manifest"
 
+local LUAUI_WIDGETS_DIR = "LuaUI/Widgets"
+local LUAUI_WIDGET_BACKUP_DIR = "LuaUI/WidgetBackups"
+local LUAUI_CONFIG_DIR  = "LuaUI/Config"
+local LUAUI_WIDGET_CONFIG_FILES = {
```
AntlerForce, line 39: These files (IGL_...) don't actually exist in this location for me.

AntlerForce:
Screenshot_20240885

This button clips into the search bar at ~768p vertical resolution.

screen_2026-08-02_22-18-49-217

Reading the PR description and comments makes it seem to me like this is aiming to reset all custom widget configuration but this didn't happen when I tested. I tried installing the Energy Conversion Meter widget, moving it's position, and then exiting the battle, using the Reset to Default button, installing the widget again, and then loading a game again. The widget was still located in my previously set custom location

Screenshot_20240887

If I install a widget, Reset to Default, then install the same widget again, I get a couple of Download Failed notifications but the widget installs anyways.

> Reset to Default no longer deletes LuaUI/Widgets instead, the existing LuaUI/Widgets directory is moved to LuaUI/WidgetBackups/Widgets_backup_<timestamp>. Then a new empty LuaUI/Widgets directory is created.

I'm unsure if this is still too risky because it's just moving the entire folder without considering what is inside it. Out of curiosity I put a bunch of video files into the /Widgets/ folder to see if they would be moved as well and they were.

ZainMGit:
Addressed the comments made above, should be ready for another review. 


Full reset. 
- The reset now clears another config file.
All removed configs; 
   `LuaUI/Config/BYAR.lua`
   `LuaUI/Config/BYAR_order.lua`
   `LuaUI/Config/BYAR_known.lua`
   `LuaUI/Config/BYAR_data.lua`
- Tested with the energy conversion widget, no longer keeps its moved position after reinstalling the widget after a reset 


Notification fix
- Hidden download failures do not trigger a user notification 
- visible install/download failures still notify normally.

UI clipping fix:
- The header button label was shortened from to `Defaults`.

AntlerForce:
`The header button label was shortened from to Defaults.`
That doesn't fix the issue.
Screenshot_20240898
Elements should be programmed not to clip into each other when there isn't enough space, in this case shortening the search bar is probably the easiest path. See the Search bar in the lobby browser for an example of this.

`The reset now clears another config file.
All removed configs;`

Again, the `BYAR_order/known/data` files don't even exist here, why are we clearing them?

WybrenKoelmans:
This button should not be in the plugins tab but in the settings tab.

- its unlikely many users will need it/use it.
- it resets ALL settings, not just plugins.

AntlerForce:
> This button should not be in the plugins tab but in the settings tab.
> 
> * its unlikely many users will need it/use it.
> * it resets ALL settings, not just plugins.

Yes, I see now that my legacy build menu has been reset to grid. This is doing much more than just resetting 
plugins.

This may not work from settings menu though, multiple times I have tried to do stuff in there (delete account, reload modoptions) and it just doesn't work but there are no significant errors in log.

ZainMGit:
My intention was to create a full "defaults" button not just a button which reset plug ins I wanted a full reset available in game which resets  all widgets back to the original state which occurs when you first download the game.

I guess if the regular ``setttings`` tab would be more appropriate for this type of reset it can be moved there, however if Moose  is correct and it will not work from the settings menu im unsure of where else it should be moved to. 


Also I corrected the button clipping by taking your advice.
Alpha: 
image
Changed:
image

PR #1234, opened 2026-07-22
various emoji changes

ZainMGit:
Few optimizations to how emojis work to remove chobby stutters. Not sure if all of these were necessary but im bundling a few optimizations into one PR. Spectated an 8v8 match and waited for catchup It "feels" faster with these optimizations so it should fix the problem.

Main problem diagnosed by Rysica: https://discord.com/channels/549281623154229250/1528864036618240140/1529352967226917106
TLDR: ordinary non emoji messages to go through emoji parsing, and each appended line rebuilt the layout for the entire chat history. Together this led to chobby taking 10+ seconds to catchup after a long 8v8 game

Changes:
1) The chat emoji detection now runs on the raw message before the timestamp and formatting is added (completes the check before the message is formatted like this "[12:00] Username: message". Before 

2) Updated `EmojiTextBox:AddLine` to append a new line instead of rebuilding the chat history

3) Limited emoji chat drawing to visible physical lines inside the scroll panel

AI Disclosure:  used Codex GPT 5.5 to assist with programming

### DIFF
--- a/LuaMenu/widgets/chobby/components/chat_emojis.lua
+++ b/LuaMenu/widgets/chobby/components/chat_emojis.lua
@@ -50,12 +50,10 @@ end
 table.sort(ChatEmojis.sortedAliases)
 
 ChatEmojis._aliasImagePath = {}
-ChatEmojis._unicodeStartChars = {}
-ChatEmojis._unicodeStartSet = {}
 ChatEmojis._candidateCache = {}
 ChatEmojis._candidateCacheCount = 0
 ChatEmojis._candidateCacheMax = 4096
-
+-- Only known :alias:  should enable emoji parsing
 local function HasRenderableAlias(text)
 	for alias in string.gmatch(text, ":([%a_]+):") do
 		if ChatEmojis._aliasImagePath[alias] ~= nil then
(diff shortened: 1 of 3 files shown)


LuaMenu/widgets/chobby/components/console.lua
```diff
@@ -476,9 +476,14 @@ function Console:AddMessage(message, userName, dateOverride, color, thirdPerson,
 		end
 	end
 
+	-- Check before formatting adds colons (raw message)
+	local allowEmoji = false
+	if userName ~= nil and ChatEmojis and ChatEmojis.HasEmojiCandidate then
+		allowEmoji = ChatEmojis.HasEmojiCandidate(message)
+	end
```
efrec: You can avoid the statement with an expression:
```suggestion
	local allowEmoji = userName ~= nil and ChatEmojis and ChatEmojis.HasEmojiCandidate and ChatEmojis.HasEmojiCandidate(message)
```

LuaMenu/widgets/chobby/components/chat_emojis.lua
```diff
@@ -80,16 +80,8 @@ function ChatEmojis.HasEmojiCandidate(text)
...
-		hasEmoji = true
-	else
-		for i = 1, #ChatEmojis._unicodeStartChars do
-			if string.find(text, ChatEmojis._unicodeStartChars[i], 1, true) then
-				hasEmoji = true
-				break
-			end
-		end
+	if string.find(text, ":", 1, true) then
+		hasEmoji = HasRenderableAlias(text)
```
efrec: Same here, would prefer an expression:

```suggestion
	local hasEmoji = string.find(text, ":", 1, true) and HasRenderableAlias(text)
```

efrec (approved):
A couple nitpicks but I'm happy with the perf. LGTM.

PR #1237, opened 2026-07-28
Add Out of range battle list filter

Rysicaa:
Adds a filter for min/max os and min/max chev to chobby battle lobby list.  It parses the names of the lobbies to figure out which ones match your criteria. 

image

For context im 6 chev ~35 os

without filter on 
image
with filter on 
image

AI used: cursor to understand/write, manual review & testing


AntlerForce:
Screenshot_20240888

At ~1080p vertical resolution the Passworded text clips into the checkbox, and it generally appears misaligned, both vertically and horizontally, and the spacing is very uneven looking.

Screenshot_20240889

At ~768p it becomes worse, though it is not great on Alpha either:

Screenshot_20240890

Rysicaa:
ack, will go through all the resolutions on another pass. will be out for a few weeks so will address comments after

Rysicaa:
image
image
image
image

Redid how the bar resizes and attached the elements to the bar itself rather than to the scrollview. This should make it nice at every resolution and improves on how it is on alpha.

LuaMenu/widgets/chobby/components/battle/battle_list_window.lua
```diff
@@ -13,6 +13,98 @@ local function ChobbyReady()
...
+
+function BattleListWindow:GetFilterRowHeight()
+	local font = Configuration:GetFont(2)
+	local fontSize = font.size or 16
+	return math.max(FILTER_ROW_HEIGHT, math.ceil(fontSize * 1.3), FILTER_BOX_SIZE + 12)
+end
+
+function BattleListWindow:GetFilterBoxSize()
+	return FILTER_BOX_SIZE
+end
```
burnhamrobertp: Why do we have a function defined which returns a variable? The only place I noticed this function used was when it was called to assign its value to yet a different local variable?

LuaMenu/widgets/chobby/components/battle/battle_list_window.lua
```diff
@@ -179,20 +290,43 @@ function BattleListWindow:init(parent)
...
+	end, 0)
+
+	self.onUiScaleChange = function ()
+		-- SetUiScale notifies listeners before screen0:Resize; defer so layout
+		-- uses the updated logical view size instead of the previous scale's.
+		WG.Delay(function ()
+			self:LayoutFilterBar()
+		end, 0)
+	end
+	Configuration:AddListener("OnUiScaleChange", self.onUiScaleChange)
```
burnhamrobertp: Can this be located with the other AddListener calls?

LuaMenu/widgets/chobby/components/battle/battle_list_window.lua
```diff
@@ -279,6 +413,9 @@ function BattleListWindow:RemoveListeners()
 	lobby:RemoveListener("OnLeftBattle", self.onLeftBattle)
 	lobby:RemoveListener("OnUpdateBattleInfo", self.onUpdateBattleInfo)
 	lobby:RemoveListener("OnBattleIngameUpdate", self.onBattleIngameUpdate)
+	if self.onUiScaleChange then
+		Configuration:RemoveListener("OnUiScaleChange", self.onUiScaleChange)
+	end
```
burnhamrobertp: Does this need to be conditional? We're not conditionally binding the listener in the first place and onUiScaleChange looks to always be defined?

LuaMenu/widgets/chobby/components/battle/battle_list_window.lua
```diff
@@ -727,6 +864,86 @@ function BattleListWindow:AddBattle(battleID, battle)
...
+	if not me then
+		return nil, nil
+	end
+	local myRating
+	if me.accountID and WG.UserHandler and WG.UserHandler.GetSnapshotSkillValue then
+		myRating = WG.UserHandler.GetSnapshotSkillValue(me.accountID, battle)
+	end
+	if not myRating then
+		myRating = tonumber(me.skill)
+	end
```
burnhamrobertp: I genuinely have no idea, but it's this not accessible through a single, reliable property?

burnhamrobertp (changes requested):
Some potential simplifications, not sure all are definitely possible. Comments, naming and structure all look good to me, layout and element code looks good but I've not much experience there.

LuaMenu/widgets/chobby/components/battle/battle_list_window.lua
Rysicaa: I went ahead and removed the check on me.skill. 

Technically GetSnapshotSkillValue is more accurate as it filters by type (duel, ffa, etc), but it does fail for some edge cases (new player, etc). In those cases, me.skill is better, but even that doesn't cover all the cases since it just is the OS from the last played game. Unfortunately afaik there's no single check to get it accurately here. 

The GetSnapshotSkillValue should cover the vast majority of cases/players though so I went ahead and removed the other one for clarity.

Rysicaa:
>  layout and element code looks good but I've not much experience there.

tbh, that makes both of us! I dabble in frontend and elements only when absolutely needed :')

LuaMenu/widgets/chobby/components/battle/battle_list_window.lua
```diff
@@ -727,6 +854,83 @@ function BattleListWindow:AddBattle(battleID, battle)
...
+end
+
+local function GetMyOpenSkillRating(battle)
+	local me = lobby:GetUser(lobby:GetMyUserName())
+	if not me then
+		return nil, nil
+	end
+	local myRating
+	if me.accountID and WG.UserHandler and WG.UserHandler.GetSnapshotSkillValue then
+		myRating = WG.UserHandler.GetSnapshotSkillValue(me.accountID, battle)
```
burnhamrobertp, line 902: Teiserver's `game_type` has a Team FFA case that `ResolveSnapshotColumns` doesn't, and `teamSize`/`nbTeams` only reach us when `lobby.Broadcast Battle Teams Information` is on (defaults off), so multi-team lobbies - and possibly most lobbies - end up compared against the Large Team column instead of the one the server actually gates on.

LuaMenu/widgets/chobby/components/battle/battle_list_window.lua
```diff
@@ -13,6 +13,93 @@ local function ChobbyReady()
...
+			rowCount = rowCount + 1
+			y = FILTER_BAR_PADDING_TOP + row * rowHeight
+			x = labelWidth
+		end
+
+		item:SetPos(x, y, itemWidth, rowHeight)
+		x = x + itemWidth + FILTER_ITEM_GAP
+	end
+
+	local barHeight = FILTER_BAR_PADDING_TOP + FILTER_BAR_PADDING_BOTTOM + rowCount * rowHeight
```
burnhamrobertp: Looks like this doesn't account for the bar's own 5px Chili padding, so the last row comes out 4px taller than the client area - harmless as it stands since the checkbox is centred, but it'd start clipping if the row height or valign ever changed.

LuaMenu/widgets/chobby/components/battle/battle_list_window.lua
```diff
@@ -279,6 +405,7 @@ function BattleListWindow:RemoveListeners()
 	lobby:RemoveListener("OnLeftBattle", self.onLeftBattle)
 	lobby:RemoveListener("OnUpdateBattleInfo", self.onUpdateBattleInfo)
 	lobby:RemoveListener("OnBattleIngameUpdate", self.onBattleIngameUpdate)
+	Configuration:RemoveListener("OnUiScaleChange", self.onUiScaleChange)
 	lobby:RemoveListener("OnConfigurationChange", self.onConfigurationChange)
```
burnhamrobertp: Unrelated to your change, but this one looks like it never removes anything - it's added on `Configuration` rather than `lobby` and `self.onConfigurationChange` is never set, so it's passing nil; your new line above it has the right shape.

LuaMenu/widgets/chobby/components/battle/battle_list_window.lua
```diff
@@ -727,6 +854,83 @@ function BattleListWindow:AddBattle(battleID, battle)
...
+	end
+	local t = string.lower(title)
+	local limits = {}
+
+	limits.minChev = tonumber(t:match("min%s*chev%s*:?%s*(%d+)"))
+	limits.maxChev = tonumber(t:match("max%s*chev%s*:?%s*(%d+)"))
+	limits.minRating = tonumber(t:match("min%s*rating%s*:?%s*(%-?%d+%.?%d*)"))
+	limits.maxRating = tonumber(t:match("max%s*rating%s*:?%s*(%-?%d+%.?%d*)"))
+
+	-- Bare "Rating lo - hi", skipping matches that are part of "min rating" / "max rating"
```
burnhamrobertp: I don't think the min/max skip can ever trigger - teiserver only writes a range in the bare `Rating: MIN-MAX` form, never after `Min rating:`/`Max rating:` - so this could probably just be one `string.find` with no loop.

LuaMenu/widgets/chobby/components/battle/battle_list_window.lua
```diff
@@ -13,6 +13,93 @@ local function ChobbyReady()
...
+local FILTER_BAR_PADDING_TOP = 2
+local FILTER_BAR_PADDING_BOTTOM = 6
+local FILTER_LIST_GAP = 4
+local FILTER_COMBO_WIDTH = 85
+local FILTER_BOX_SIZE = 20
+
+function BattleListWindow:GetFilterRowHeight()
+	local font = Configuration:GetFont(2)
+	local fontSize = font.size or 16
+	return math.max(FILTER_ROW_HEIGHT, math.ceil(fontSize * 1.3), FILTER_BOX_SIZE + 12)
```
burnhamrobertp: This always comes out 32 - font 2's size is hardcoded at 17 and never changes, so `FILTER_ROW_HEIGHT` and the font term can never win the max.

LuaMenu/widgets/chobby/components/battle/battle_list_window.lua
```diff
@@ -13,6 +13,93 @@ local function ChobbyReady()
...
+	local rowHeight = self:GetFilterRowHeight()
+	local labelWidth = font:GetTextWidth(self.filterLabel.caption) + FILTER_ITEM_GAP
+	local x = labelWidth
+	local y = FILTER_BAR_PADDING_TOP
+	local row = 0
+	local rowCount = 1
+
+	for i = 1, #self.filterItems do
+		local item = self.filterItems[i]
+		if item.boxsize and item.boxsize ~= FILTER_BOX_SIZE then
```
burnhamrobertp: This never fires - every item is already built with `boxsize = FILTER_BOX_SIZE`, and the skin's 13 can't override an explicit value because `table.merge` only fills in nils.

burnhamrobertp (changes requested):
Went through the filter against teiserver's title generator. The chev side looks right to me. One on the rating side that I think is a real behaviour gap, plus some dead code in the layout.

LuaMenu/widgets/chobby/components/battle/battle_list_window.lua
Rysicaa: > teamSize/nbTeams only reach us when lobby.Broadcast Battle Teams Information is on (defaults off)

I actually checked this. It looks like we are getting the teamSize / nbTeams info in chobby. So it appears that in the configuration for teiserver, the broadcast information is indeed on

Rysicaa: We can also do this by parsing the 5v5, 2v2, 2v2v2, etc, but it did seem cleaner to just use the existing info from the broadcast

LuaMenu/widgets/chobby/components/battle/battle_list_window.lua
Rysicaa: Good catch. The one below it, DownloadFinished, actually seems to be the same.

LuaMenu/widgets/chobby/components/battle/battle_list_window.lua
```diff
@@ -727,6 +848,70 @@ function BattleListWindow:AddBattle(battleID, battle)
...
+local function ParseJoinLimitsFromTitle(title)
+	if not title or title == "" then
+		return nil
+	end
+	local t = string.lower(title)
+	local limits = {}
+
+	limits.minChev = tonumber(t:match("min%s*chev%s*:?%s*(%d+)"))
+	limits.maxChev = tonumber(t:match("max%s*chev%s*:?%s*(%d+)"))
+	local lo, hi = t:match("rating%s*:?%s*(%-?%d+%.?%d*)%s*%-%s*(%-?%d+%.?%d*)")
```
burnhamrobertp, line 883: `Max rating: 25` and `Min rating: 10` don't parse at all now - I ran this against every title teiserver generates and only the `Rating: MIN-MAX` form comes back with anything, where both single-bound forms worked before this push. They're definitely reachable: `!maxratinglevel` and `!minratinglevel` each leave the other bound alone, so a lobby with only one of them set gets the short form in its title - and going by teiserver's own noob-lobby tips, `!maxratinglevel` is the usual way those get set up, so it's probably the common case rather than an edge one. Think you want to keep both `t:match` lines and only drop the loop that was around them.

LuaMenu/widgets/chobby/components/battle/battle_list_window.lua
```diff
@@ -727,6 +848,70 @@ function BattleListWindow:AddBattle(battleID, battle)
 	self:RecalculateOrder(battle.battleID) -- when a battle is added to the list, go ahead and ensure it's sorted correctly. Other cases will rely on soft update
 end
 
+-- Parse chevron/rating join limits advertised in battle titles, e.g.
+-- "Min chev: 4 | Max chev: 6 | Rating: 13-60"
```
burnhamrobertp: Small thing, `Min chev: 4 | Max chev: 6` isn't a title teiserver can actually produce - setting either chev bound resets the other, so you only ever see one of them at a time.

LuaMenu/widgets/chobby/components/battle/battle_list_window.lua
```diff
@@ -13,6 +13,84 @@ local function ChobbyReady()
...
+			rowCount = rowCount + 1
+			y = FILTER_BAR_PADDING_TOP + row * FILTER_ROW_HEIGHT
+			x = labelWidth
+		end
+
+		item:SetPos(x, y, itemWidth, FILTER_ROW_HEIGHT)
+		x = x + itemWidth + FILTER_ITEM_GAP
+	end
+
+	local padding = self.filterBar.padding or {0, 0, 0, 0}
```
burnhamrobertp: Looks like the padding is handled twice now - the bar is built with `padding = {0, 0, 0, 0}` further down, so `padY` here is always 0; the pinned padding on its own is probably enough, and worth noting the height in the constructor doesn't add `padY` while this does, so the two would disagree on the first frame if anyone ever unpins it.

burnhamrobertp (changes requested):
Heads up, the rating parse lost the single-bound forms in this push - `Max rating: 25` and `Min rating: 10` match nothing now, where they both worked before.

Half of that is on me. My last comment said the min/max skip could never trigger, and I meant the loop's prefix guard, not the two `t:match` lines above it. Should have been clearer.

The other half is worth a word since you mentioned cursor in the description. An agent handed that comment will do exactly what it says and take the patterns out along with the loop - it has no way of knowing `!maxratinglevel` is how most noob lobbies get set up, and it isn't weighing that either way. They follow the instruction, they don't think about it. So anything one of them writes wants a read-through locally before it goes up, especially when the change is deleting something rather than adding it.

Rest of the fixes look good to me - the listener cleanup went further than I flagged, which is nice to see.

Rysicaa:
Yeah, that's totally fair. Tbh I really just shouldn't code late at night. I don't catch stuff I normally do.

Btw, on the other issue with team ffa, went down a lil rabbit hole, but I put out a PR on teiserver to use large team OS for the restrictions. https://github.com/beyond-all-reason/teiserver/pull/1495/changes

It's _technically_ dependent on it to be accurate, but realistically, so few people actually play team ffa with an os restriction that I'm pretty sure no one would notice.

burnhamrobertp:
If you can resolve feedback items that are resolved, that would be helpful;

LuaMenu/widgets/chobby/components/battle/battle_list_window.lua
```diff
@@ -727,6 +846,72 @@ function BattleListWindow:AddBattle(battleID, battle)
 	self:RecalculateOrder(battle.battleID) -- when a battle is added to the list, go ahead and ensure it's sorted correctly. Other cases will rely on soft update
 end
 
+-- Parse chevron/rating join limits advertised in battle titles, e.g.
+-- "Min chev: 4 | Max chev: 6 | "Min rating: 10" | "Max rating: 25" | Rating: 13-60"
```
burnhamrobertp: Quotes inside some of the items here are erroneous, and even the surrounding quotes aren't necessary.

burnhamrobertp (approved):

burnhamrobertp (changes requested):
The screenshots in the body are no longer accurate, and the current version of the UX is a little worse as a result. The spacing along the bottom filter bar is now poorly utilized with all filters clumped on the left hand side.

Rysicaa:
The screenshots from this one are accurate: https://github.com/beyond-all-reason/BYAR-Chobby/pull/1237#issuecomment-5338865613 

Just to clarify @burnhamrobertp, what were you thinking? equally spaced? A little more margin/padding between them? 

It really just depends on the resolution and since the UI scale is a slider, there's always some resolution where it's subjectively bad. For example, if we did equally spaced, it would look quite bad on larger screens/resolutions as there would be huge gaps between options. 

I don't have strong opinions on this. Just let me know what you prefer and we can change it to that. 

I would also say that since this is already scope crept beyond the original scope to fix clipping here, I'm a little less inclined to make drastic changes or rehaul the implementation of it to do something like dynamic scaling at every resolution.

burnhamrobertp:
Equally spaced seems a better use of the area. That's the current implementation on master, as well, when viewing the lobby at a large enough resolution that there is unused space.

Rysicaa:
Ok, we can do that. We'll equally space it and go to next row if there is not enough space.

Rysicaa:
image
image
image
image
image

burnhamrobertp:
When testing this, there's a period of 30s - 1 minute after coming back to the lobby where the "Out of Range" filters don't actually filter out lobbies that are out of range. The underlying cause for this needs to be identified and addressed. For me, I saw it at least with Max rating for a lobby being below my threshold, I believe for a large team game.

Rysicaa:
> When testing this, there's a period of 30s - 1 minute after coming back to the lobby where the "Out of Range" filters don't actually filter out lobbies that are out of range. The underlying cause for this needs to be identified and addressed. For me, I saw it at least with Max rating for a lobby being below my threshold, I believe for a large team game.

This is expected as a result of chobby's existing architecture for handling player openskill info. 

When you're loading lobby for the first time your game needs to download the openskill snapshot csv. That's where all of the skill rating data is pulled from. Nothing else that relies on this data in chobby loads properly. If you try loading a player OS tooltip while it's loading, it won't show anything either. 

It doesn't happen if you return to lobby after a game, only after opening game for first time after ~24 hrs (or if cache is cleared etc).

burnhamrobertp:
Either we should be disabling that filter until its able to read from the cache, or else we bypass the loading of the entire openskill snapshot file and parse just the line for the current user?

I'd prefer the latter, but the former is - I'm sure - cheaper/easier.

Rysicaa:
image

PR #1241, opened 2026-08-11
[Draft/Suggestion] Focus replays opened by the launcher

eun-ice:
## Status

> **Draft / suggestion / base for discussion.** This is intentionally proposed as a temporary compatibility improvement for the legacy Chobby client, not as a competing direction for the new client.

The equivalent workflow is expected to arrive in the new client. Still, this small bridge may help some users already using the current launcher and Chobby while that transition is ongoing. Maintainer feedback on whether carrying this temporary solution is worthwhile is very welcome.

## What this suggests

- announce to the launcher when Chobby's replay UI is ready;
- receive an `OpenReplay` request through the existing launcher loopback;
- open the Replays screen, refresh it when necessary, and prioritize/scroll to the imported replay;
- keep replay launching itself as an explicit user action.

The paired launcher change imports the external `.sdfz` into `demos/` and sends the relative replay path after this readiness handshake.

Paired launcher PR: https://github.com/beyond-all-reason/spring-launcher/pull/54

## Verification

- Manually verified on Linux with the local `BYAR-Chobby` checkout linked as **Dev Lobby**.
- Dragging an external `.sdfz` into the paired launcher imported it, started BAR, opened the Replays screen, and focused the imported replay.
- Existing replay list usage and explicit replay start behavior remain unchanged.

## AI usage disclosure

OpenAI Codex was used interactively to trace the launcher/Chobby bridge, assist with implementation, tests, debugging, and drafting this PR text. I reviewed the changes and manually verified the end-to-end behavior described above.

### DIFF
--- a/LuaMenu/widgets/gui_chili_lobby.lua
+++ b/LuaMenu/widgets/gui_chili_lobby.lua
@@ -106,6 +106,9 @@ function widget:Initialize()
 	lobbyInterfaceHolder = interfaceRoot.GetLobbyInterfaceHolder()
 	Chobby.lobbyInterfaceHolder = lobbyInterfaceHolder
 	Chobby.interfaceRoot = interfaceRoot
+	if WG.WrapperLoopback then
+		WG.WrapperLoopback.ReplayHandlerReady()
+	end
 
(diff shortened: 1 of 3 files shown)


eun-ice:
closed since launcher PR not accepted.

PR #1243, opened 2026-08-21
Render polygon start box overrides

burnhamrobertp:
Follows up on #1184: setting the override by hand with !bSet using a polygon got thrown away by the lobby, so the minimap kept showing the map's default boxes while the game went ahead and enforced the polygon.


p2004a (approved):
Thanks!

PR #1245, opened 2026-08-21
Feat: development hot reload widget with shortcuts 

tetrisface:
No related github issue.

This adds a setting that will auto hot reload widgets that you are developing.

This probably doesn't work flawlessly for many widgets and there probably is a more proper reliable implementation. But I think it's worth to get the work started and this is quite small and non-invasive.
Something to think about at least.

image

Ctrl+T will toggle it and Ctrl+R will run it on demand.

Tested with the modoptions panel sort widget development and it worked good. It has a lot of custom code to make it force rerender so it's not a super good example.
I also tested the guard so it doesn't conflict with in-game keybinds.

# AI Disclosure
99% done with Claude Fable


tetrisface:
I don't think this meets the AI policy

PR #1249, opened 2026-08-24
Add a dev-only Map Editor menu entry

burnhamrobertp:
Adds a dev-only Map Editor entry to the singleplayer menu that launches straight into a terraformer session on a chosen map or a saved project, instead of starting a normal game and setting it up by hand every time.

Depends on beyond-all-reason/Beyond-All-Reason#8886. The start script sets the `mapeditor` modoption and nothing else, so without the game-side widget reading it you land in an empty spectator session with no editor at all.

The map list is its own implementation rather than a second caller of the map browser's row builder, since this is a prototype we expect to reshape. Verified by launching both a map and a saved project from the lobby.

### DIFF
--- a/LuaMenu/configs/gameConfig/byar/singleplayerMenu.lua
+++ b/LuaMenu/configs/gameConfig/byar/singleplayerMenu.lua
@@ -10,6 +10,13 @@ local items = 	{
 		--startWithTabOpen = 1,
 	},
 	{
+		-- devOnly is honoured by interface_root, which is where Configuration.devMode is
+		-- readable: this file is included while Configuration is still being built.
+		name = "terraformer",
+		control = WG.TerraformerWindow.GetControl(),
+		devOnly = true,
+	},
+	{
 		name = "load_game",
 		control = WG.LoadGameWindow.GetControl(),
 		entryCheck = WG.BattleRoomWindow.SetSingleplayerGame,
--- a/LuaMenu/widgets/chobby/components/interface_root.lua
+++ b/LuaMenu/widgets/chobby/components/interface_root.lua
@@ -510,25 +510,33 @@ function GetInterfaceRoot(optionsParent, mainWindowParent, fontFunction)
 		ToggleShowFunc(obj, newTab)
 	end
 
+	-- Built up rather than written as one literal: the singleplayer tabs were indexed by hand,
+	-- so adding one meant editing here too, and devOnly needs checking somewhere Configuration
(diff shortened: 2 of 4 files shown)


AntlerForce:
This didn't work for me, I tried loading a terraformer project with this interface but something was wrong with the resulting heightmap, it had a large blank flat space on the left side that isn't present when loading with the ingame interface.

Texture didn't match what I get from an ingame load either.

If this did work, it would be more useful if the project list was sorted by most recently updated, rather than name.

burnhamrobertp:
Are you able to send me that terraformer project (or provide steps for me to reproduce this myself) to make sure I've not missed something in the startscripts that are generated in chobby for this? I can change the project list ordering, that's not a big deal, though it might require a tweak in game code to ensure that some form of last updated on is written back out somewhere, for each project.

AntlerForce:
> Are you able to send me that terraformer project (or provide steps for me to reproduce this myself) to make sure I've not missed something in the startscripts that are generated in chobby for this? I can change the project list ordering, that's not a big deal, though it might require a tweak in game code to ensure that some form of last updated on is written back out somewhere, for each project.

The heightmap bug is easy to reproduce.
(1)Open great divide
(2)save project from terraform brush.
(3)Close completely and start game from launcher to get the project to appear in chobby menu
(4)Open saved project from chobby

The result is this:
screen_2026-09-03_19-53-57-450

burnhamrobertp:
A handful of other obvious UX changes / improvements made as well

burnhamrobertp:
This is waiting for the changes in https://github.com/beyond-all-reason/Beyond-All-Reason/pull/8886 to make their way to stable

PtaQQ:
One request: make the projects browser also check folders in the MapProjects folder. with 1.14 I added folder tree support. 

image

text alignment needs a looksie here: 
image

For future - I will add a minimap img to the map projects so you can also show previews there. 

Otherwise all seems good.

PtaQQ:
Wait I was able to also reproduce the bug Moose noticed, seems like its still there: 
image

1. open an existing map project 
2. quit game 
3. without selecting any map project click start again 
this 

I don't see a way to start a completely new map as well, since even without selection it starts avalanche for me. I think there should be a separate explicit New Map button for clarity. 

small ui nitpick:
image

after the game has started this stays saying "starting" should turn to "running" or "reload" or something.

burnhamrobertp:
For now, the directory support works, but is crude.
image

PR #1252, opened 2026-09-03
Make the battle room controls match the room's actual settings

burnhamrobertp:
Replaces the Add Team button with a team count selector wired to nbTeams, and moves team size and lock state onto controls that show what the room actually has.

Add Team never told the server anything. It built a team panel locally, and since you cannot start without calling !balance, and balance drops any allyteam above nbTeams, that team was gone before the game began. So going through !nbTeams does not add a vote that was not already there. Same reason an emptied team now stays put: the count decides how many teams a room has, not who is standing in them.

The !balance that follows a count or size change waits for the line SPADS prints naming whoever asked for it, since sending it next to the command balances the old shape while the setting is still sitting in a vote.

Start boxes are in here because the count picks which arrangement renders. Removing the last box used to clear the override and quietly restore the map defaults, so it now records an override of no boxes, and Start refuses a room where a team has none. Custom Boxes is selectable rather than only a label because adding a box already switched the room onto an override of its own, so picking it directly grants nothing new. The selector takes its value from whatever override the room holds, so it follows a change somebody else makes rather than only the ones made here.

On a map with polygon start boxes, picking Custom Boxes looks like it sends an override of no boxes rather than freezing the polygons into rects you can drag, because the overlay never fills in the rect values that path copies from. Left alone here.


burnhamrobertp:
Able to have no startboxes (this seems weird but the current behavior is far worse / more unexpected)
2026-09-03 13_15_59-Clipboard

No more having "Default Startboxes" show up when you have custom startboxes defined
2026-09-03 11_22_52-Clipboard

nbTeams and teamsize moved up to here, "Add team" button is removed, everything stays in sync automatically
2026-09-03 13_00_41-Clipboard

Reduced number of things in this part of the lobby UI, including removed "Locked" select in favor of just using the buttons we have that people like, but now they also show your current state.
2026-09-03 13_01_09-Clipboard

